PR #45, opened 2015-04-01
Add pickling support (fixes #42)

### LINKED ISSUE #42: pickle causes Immediate segfault
Hi, if I run the following:
```
import rtree
import cPickle
x = rtree.index.Index()
cPickle.loads(cPickle.dumps(x))
```
I get a segfault when I quit python, (or at another random time later, depends how long it runs for). I've tried using plain 'pickle' as well, but get the same behaviour.

I'm using rtree 0.8.2 (installed via pip) against libspatialite 4.2.0-2 on arch linux x64, python 2.7.9

kilink:
Make Index and Property objects pickleable. This fixes issue #42 .

Also, move ctypes handle objects into their own wrapper objects. The lifetime of the underlying C object is tied to the wrapper. This makes it easier to pass handles without worrying about ownership.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,4 +15,4 @@ install:
   - pip install -e .
 
 script:
-  - py.test tests/test_doctests.py tests/test_index.py
+  - py.test tests/test_doctests.py tests/test_index.py tests/test_pickle.py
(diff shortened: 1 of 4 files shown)


hobu:
Awesome! Thanks so much.

Issue #56, opened 2015-09-29
ctypes find_library does not find libspatialindex_c.so in non-standard locations


samfux84:
Please do not use find_library from ctypes to find libspatialindex_c. This only works if it is installed on in a standard location. On HPC clusters, it is very likely that the library is not installed in a standard location. Adding the installation directory to LD_LIBRARY_PATH, or specifying the installation directory in the setup.cfg file should be sufficient, but it does not work.

The only way to make rtree find libspatialindex_c on a Linux HPC cluster is to "hack" core.py, comment out the line

lib_name = find_library('spatialindex_c')

and replace lib_name in

rt = ctypes.CDLL(lib_name)

with the absolut path of the library.

I understand that the problem is introduced by ctypes and not by the rtree package, but please use another mechanism than find_library to detect libspatialindex_c on Linux systems. The two most practical options would be to either use LD_LIBRARY_PATH or the setup.cfg file.

Thank you very much and best regards

Samuel Fux

hobu:
What is the Python Way™ to do this nowadays? The ctypes find_library code is from eons ago (taken from Shapely which took it from something else IIRC).

We need something that doesn't interfere with virtualenv and installs via pip to the right location.

samfux84:
In my opinion, it is still ok to use ctypes find_library (it seems to work on all operating systems except Linux), but if find_library returns nothing, then it would be nice if there was a fallback option that opens the setup.cfg file and checks if the path to libspatialindex_c is specified there.

hobu:
Please make a pull request implementing as you propose.

mwtoews:
I can confirm that if spatialindex is build with `./configure --prefix=/some/dir`, and `/some/dir` is on `LD_LIBRARY_PATH`, then `ctypes.CDLL('libspatialindex_c.so')` correctly finds the library, while `find_library` does not work (returns None). Perhaps [these lines in core.py](https://github.com/Toblerity/rtree/blob/6f2e78454b567cc70e5c3d09caa3a5373c508656/rtree/core.py#L103-L109) should be changed:
```diff
@@ -103,5 +103,3 @@
 elif os.name == 'posix':
-    platform = os.uname()[0]
-    lib_name = find_library('spatialindex_c')
-    rt = ctypes.CDLL(lib_name)
+    rt = ctypes.CDLL('libspatialindex_c.so')
 else:
```

samfux84:
[...]

PR #74, opened 2016-05-29
Bake in env vars

citibeth:
When env SPATIALINDEX_LIBRARY or SPATIALINDEX_C_LIBRARY are set during setup time, bake them into the installation so they do not have to be set up at run time.  This resolves the problems I had in #56.

**To be clear:** If you do NOT set these env vars at build time, then rtree will work exactly as before: you will be able to set them at runtime.  Currently, run-time settings of these env vars do NOT override build-time settings.

Further rational for why this change is needed... I use Spack to build my software stack ( https://github.com/llnl/spack ).  How will rtree find libspatialindex?  There are a few ways this might happen:
1. Normally, Spack builds libraries with RPATH built-in.  Binaries can find their dependencies without the user having to set LD_LIBRARY_PATH.  This is really useful because it's easy to set LD_LIBRARY_PATH wrong.  Or to have binary A require one LD_LIBRARY_PATH and binary B require a different LD_LIBRARY_PATH --- and you want to be able to run them both from the same command shell.
   RPATH doesn't work here because the library is being loaded by Python code, not the system loader.
2. Spack produces an environment module for each package, which lets the package export stuff about itself.  One could upgrade the Spack package for libspatialindex to export the required env vars.  However... why should libspatialindex have to do this, just for the convenience of rtree?  This violates a principle of locality.  Rtree should be able to figure out what it needs at build time without any special help from libspatialindex (Spack already knows where libspatialindex is located at build time).

Rather than (2), it's better to modify rtree as appropriate, so libspatialindex doesn't have to set any magic env vars.  This PR does that modification, implementing a system that's functionally to the "gold-standard" RPATH described in (1).

See here for the Spack recipe for rtree.  Note how the recipe sets SPATIALINDEX_LIBRARY and SPATIALINDEX_C_LIBRARY, which get baked into the rtree install:
https://github.com/citibeth/spack/blob/efischer/develop/var/spack/repos/builtin/packages/py-rtree/package.py

### DIFF
--- a/rtree/core.py
+++ b/rtree/core.py
@@ -1,5 +1,6 @@
 import os
 import ctypes
+import pkgutil
 from ctypes.util import find_library
 
 
@@ -68,6 +69,16 @@ def free_error_msg_ptr(result, func, cargs):
     p = ctypes.cast(result, ctypes.POINTER(ctypes.c_void_p))
     rt.Index_Free(p)
     return retvalue
+
+# Read back env vars that were set up during package installation.
+try:
+    for line in pkgutil.get_data('rtree', 'ENVIRON.txt').decode().split():
+        kv = line.split('=')
+        if len(kv) == 2:
+            os.environ[kv[0]] = kv[1]
+except IOError:
+    # The data file doesn't exist; that is OK
+    pass
 
 
 if os.name == 'nt':
--- a/rtree/core.py
+++ b/rtree/core.py
@@ -73,11 +73,25 @@ def free_error_msg_ptr(result, func, cargs):
 # Read back env vars that were set up during package installation.
 try:
     for line in pkgutil.get_data('rtree', 'ENVIRON.txt').decode().split():
+        # Ignore lines with wrong number of '=' signs
         kv = line.split('=')
-        if len(kv) == 2:
-            os.environ[kv[0]] = kv[1]
+        if len(kv) != 2:
+            continue
+        key, val = kv
+
+        # Only allow certain vars to be set
+        legal_envs = ('SPATIALINDEX_LIBRARY', 'SPATIALINDEX_C_LIBRARY')
+        if key not in legal_envs:
+            raise ValueError(
+                'Illegal variable in ENVIRON.txt: %s.  Must be one of %s' %
+                (key, str(legal_envs)))
+
+        # Only set from config file if not already set in env
+        if key not in os.environ:
+            os.environ[key] = val
+
 except IOError:
-    # The data file doesn't exist; that is OK
+    # The data file ENVIRON.txt doesn't exist; that is OK
     pass
 
 
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 from setuptools import setup
 from setuptools.command.build_py import build_py
 
@@ -13,7 +12,7 @@ class rtree_build_py(build_py):
             target_dir = self.build_lib
 
             # mkpath is a distutils helper to create directories
-            self.mkpath(target_dir)
+            self.mkpath(os.path.join(target_dir, 'rtree'))
 
             with open(os.path.join(target_dir, 'rtree/ENVIRON.txt'), 'w') as fobj:
                 if 'SPATIALINDEX_LIBRARY' in os.environ:
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,29 @@
 #!/usr/bin/env python
 from setuptools import setup
+from setuptools.command.build_py import build_py
 
 import rtree
+
+# Transfer key env vars to a Python data file.
+# http://www.digip.org/blog/2011/01/generating-data-files-in-setup.py.html
+class rtree_build_py(build_py):
+    def run(self):
+        # honor the --dry-run flag
+        if not self.dry_run:
+            target_dir = self.build_lib
+
+            # mkpath is a distutils helper to create directories
+            self.mkpath(target_dir)
+
+            with open(os.path.join(target_dir, 'rtree/ENVIRON.txt'), 'w') as fobj:
+                if 'SPATIALINDEX_LIBRARY' in os.environ:
+                    fobj.write('SPATIALINDEX_LIBRARY=%s\n' % os.environ['SPATIALINDEX_LIBRARY'])
+                if 'SPATIALINDEX_C_LIBRARY' in os.environ:
+                    fobj.write('SPATIALINDEX_C_LIBRARY=%s\n' % os.environ['SPATIALINDEX_C_LIBRARY'])
+
+        # distutils uses old-style classes, so no super()
+        build_py.run(self)
+
 
 # Get text from README.txt
 with open('docs/source/README.txt', 'r') as fp:
@@ -21,6 +43,7 @@ else:
     data_files = None
 
 setup(
+    cmdclass      = {'build_py' : rtree_build_py},
     name          = 'Rtree',
     version       = rtree.__version__,
     description   = 'R-Tree spatial index for Python GIS',


sgillies:
@citibeth I can't think of a reason why we should not do something like this. On the other hand I'd like to propose some modifications.

The `ENVIRON.txt` file seems a lot like a configuration file, and generally environment variables override file configuration, yes? So I'd prefer that we read the config file then overlay environment variables on its values instead of reading the config directly into the environment. I'm also a bit concerned about letting the config file carry any environment variable at all. It seems too powerful a feature. Let's limit it to just the configuration of spatialindex library paths.

@hobu how does that sound to you?

@citibeth thanks for the PR and sorry for the slow response on my part!

citibeth:
OK, thanks for looking at it.  I just pushed a commit that addresses your comments.  I haven't tried running this, but the code looks pretty good to me.  I'll see if I can try running later today.

citibeth:
OK, I tested it out on a full install, everything works great!

Please tell me if you intend to make a release with this patch soon, or if you just want to merge the PR and wait for the next release some time in the future.  That will affect what I do on my end.  See:

https://github.com/LLNL/spack/pull/1548

rtree/core.py
```diff
@@ -69,6 +70,30 @@ def free_error_msg_ptr(result, func, cargs):
...
+        # Only allow certain vars to be set
+        legal_envs = ('SPATIALINDEX_LIBRARY', 'SPATIALINDEX_C_LIBRARY')
+        if key not in legal_envs:
+            raise ValueError(
+                'Illegal variable in ENVIRON.txt: %s.  Must be one of %s' %
+                (key, str(legal_envs)))
+
+        # Only set from config file if not already set in env
+        if key not in os.environ:
+            os.environ[key] = val
```
sgillies: Yes, this gets the precedence right, but instead of using `os.environ` as data structure in this module (bc it's a global) can you implement something like the following instead?

```
libname = 'spatialindex_c.dll'
if 'SPATIALINDEX_C_LIBRARY' in config_file_dict:
    libname = config_file_dict['SPATIALINDEX_C_LIBRARY']
if 'SPATIALINDEX_C_LIBRARY' in os.environ:
    libname = os.environ['SPATIALINDEX_C_LIBRARY']

lib_path, lib_name = os.path.split(libname)
rt = _load_library(lib_name, ctypes.cdll.LoadLibrary, (lib_path,))
```

It's better if we don't update the environment unnecessarily and only access it once.

citibeth:
I agree, that's a better way to go.  Would you be able to make that
modification to the PR?  I've added you as a collaborator on my fork.

hobu:
Just a nag that I finally figured out #77 which is what was holding up having another Rtree release. I'd like to kick out another release here pretty quickly now.

I haven't been tracking this, but I have a couple of questions:

* Is this the solution for Fiona/Shapely/Rasterio for this issue too? I'd rather not duplicate code if possible.
* This approach adds a `pkgutil` dependency that wasn't there before, correct?

Howard

citibeth:
>     Is this the solution for Fiona/Shapely/Rasterio for this issue too? I'd rather not duplicate code if possible.

No idea.  Never heard of Fiona/Shapely/Rasterio

>    This approach adds a pkgutil dependency that wasn't there before, correct?

It shouldn't.  I don't see any such dependencies added in the PR.

Sorry, I'm confused at this point.  Please tell me if you are expecting me to make the suggested change above, or if you are going to do it?  I would very much like to get the functionality of this PR into the next release.

hobu:
@citibeth sorry that chatter was pointed at @sgillies. I'm 👍 on merging something like this, but I want to be convinced that the pattern it uses is used in other projects, doesn't have any bad consequences for allowing people to override behavior, and isn't going to break normal installs.

https://github.com/Toblerity/rtree/pull/74/commits/bb1ae216ef2927816af4d475a240cc7089f7426c#diff-7b81c0efed48af35e564faba8dbc6cd5R3 added the `pkgutil` dependency. Is that not still active?

Another question. If libspatialindex did RPATH appropriately, would this issue go away? I am hoping to sync a new libspatialindex release soon and if there were PRs that improved its RPATH situation, I'd happily merge them.

citibeth:
> @citibeth sorry that chatter was pointed at @sgillies. I'm 👍 on merging something like this, but I want to be convinced that the pattern it uses is used in other projects

Most Python interface projects I've seen use some kind of C-based extension to interface between the worlds of Python and C -- or if not a hand-built C extension, something with Pyhthon, SWIG or similar.  With all these approaches, one ends up with a .so files that goes in the Python install directory.  For example, the Icebin library (http://github.com/citibeth/icebin) produces a file `_icebin.so`.  Because the `.so` file is a Python extension placed in the Python install directory, Python is able to find it on the `$PYTHONPATH`.

In many/most cases, these Python extensions built on separately-built C libraries (such as `libspatialindex.so`) that reside somewhere else on the system.  If you build a C program against that library, it is best to put an RPATH into that C program so the correct library will always be picked up; without that, you can set the `$LD_LIBRARY_PATH` variable to tell the dynamic linker where to look.

So we have a "3-step" process:  Python code --> Python extension --> C library.  The Python extension serves as a "bridge" between the Python system (which doesn't know about RPATH or how to search LD_LIBRARY_PATH) and the OS (which does).  Ideally, RPATH will be used in the Python extension (we're working on making that universal over at Spack).

RTree is different because it uses CFFI, requiring the Python code to know where to find the C library directly, without the bridge.  It's the only Python extension-based package I've seen that does that, so I don't know how other projects do it.  However, setting an env var to tell the package where to find the library is clearly a hack that won't scale (imagine if every Python package needed its own env var to work).  I felt the best approach would be to get functionality similar to RPATH in the compiled-code world --- i.e. somehow "bake in" the location of the library when you install.  This PR piggy-backs on top of the existing way using the library-specific env var; had I been doing it from scratch, I would not have used an env var at all (nor would I make it easy for the user to change that setting after installation; this can only lead to things breaking).

Clearly, this approach should be extended to all Python extensions that use CFFI.  That project is beyond the scope of my interest at this point.

> doesn't have any bad consequences for allowing people to override behavior

I would say this avoids bad consequences, because it ensures the Python code will always be picking up the version of the C library it was aware of when installed.  This is not / should not be a user-settable "feature."

> bb1ae21#diff-7b81c0efed48af35e564faba8dbc6cd5R3 added the pkgutil dependency. Is that not still active?

`pkgutil` is a part of core Python, it's not an actual added dependency.

> Another question. If libspatialindex did RPATH appropriately, would this issue go away?

No.  If `libspatialindex.so` is built properly with RPATH, that means that all libraries *it* depends on will be found.  But that does nothing to help the Python code find `libspatialindex.so`.  (Spack in fact does build all binaries with RPATH, and it does not solve the problem addressed by this PR).

> I am hoping to sync a new libspatialindex release soon and if there were PRs that improved its RPATH situation, I'd happily merge them.

I would think of this PR as an RPATH-like system for Python packages that use CFFI.

sgillies:
> Clearly, this approach should be extended to all Python extensions that use CFFI. That project is beyond the scope of my interest at this point.

Adding this rpath-like feature to the ctypes or cffi modules could take some time and energy, for sure. Still, it would be nice to have this feature done well in rtree and provide a good example or even reusable code for other projects. What we have in this PR now is a good idea but code that is not easy to test and seems likely to become deprecated over time. Something that would be easy to test and more maintainable and reusable: a wrapper around ctypes `load_library()` that looks in package data for the baked-in library paths, in its own module, removing all the details from `rtree.core`.

@citibeth would such an implementation of your idea be interesting to you?

citibeth:
Sean,

I looked at the library loading stuff in RTree.  IMHO, it's a mess.  I
agree it needs refactoring, but that jobs is larger than what I really
wanted to get into, nor do I have the Windows platform to test the result.
As hackey as this PR is, I don't think it's any hackier than the code
that's already in there.  See:

                # temporarily add the path to the PATH environment variable
>                 # so Windows can find additional DLL dependencies.
>                 try:
>                     oldenv = os.environ['PATH']
>                     os.environ['PATH'] = path + ';' + oldenv


The bottom line is, `ctypes.find_library()` doesn't work.  Apparently,
people already figured that out for Windows, hence the specialized Windows
code in RTree.  But it also doesn't work for Linux.  The CTypes folk
realized this:

If wrapping a shared library with ctypes
> <https://docs.python.org/2/library/ctypes.html#module-ctypes>, it *may*
> be better to determine the shared library name at development time, and
> hardcode that into the wrapper module instead of using find_library() to
> locate the library at runtime.


If I were re-doing this stuff in RTree, I would eliminate all the dynamic
library finding out of `core.py`; instead, I RTree would just read a file
using `pkgutil` and load the library specified in that file.  All logic
required to locate libraries would be moved to `setup.py`.  But... if we
rely on users to pass a flag into `setup.py` to locate the library, then
that logic would become pretty easy.

I think we can eliminate most/all of the `if os.name == ...` stuff.

But again... this is beyond the scope of what I had hoped to do.  I just
wanted RTree to work with Spack...

-- Elizabeth

davidedelvento:
@citibeth Even though fixing this problem in cffi is beyond the scope of what interests you, I think you ought to let them know with a bug report (if you do, please link the URL of that report here).

sgillies:
I moved this one off the 0.8.3 milestone because it's more than a patch level change and needs more work.

tomplex:
Many GitHub issues link to this PR regarding a failing install with an error of `AttributeError: python3.5: undefined symbol: Error_GetLastErrorNum`. However, this PR is still open, and no one has given a concrete answer as to how I can resolve this issue and effectively install the package. I'm using a clean python3.5 virtualenv on Ubuntu 16.04. Does anyone have any advice?

hobu:
I think this story will be superseded by creating wheels that include libspatialindex, and maybe that is a better way to attack this problem.

ninnghazad:
> Many GitHub issues link to this PR regarding a failing install with an error of `AttributeError: python3.5: undefined symbol: Error_GetLastErrorNum`. However, this PR is still open, and no one has given a concrete answer as to how I can resolve this issue and effectively install the package. I'm using a clean python3.5 virtualenv on Ubuntu 16.04. Does anyone have any advice?

For me it was as simple as `apt-get install libspatialindex-dev;pip install rtree;`.

PR #82, opened 2016-12-15
Include all text files in sdist

sebastic:
The `CREDITS.txt` file is not included in the tarballs on PyPI.


hobu:
Bas, I will not issue a new release for this unless you find something else significant in your packaging activities.

sebastic:
No need for a new release, I just noticed this issue when I compared the PyPI tarball content with the git repository.

PR #102, opened 2018-04-09
Use Read the Docs

### LINKED ISSUE #84: Documentation is outdated
http://toblerity.github.com/rtree/ (in the info here) redirects to http://toblerity.org/rtree/ which is the documentation for 0.7.0, when there have been 4 releases after that.

Juanlu001:
Fix #84

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ addons:
       - libspatialindex-dev
 
 install:
+  - pip install pytest
   - pip install -e .
 
 script:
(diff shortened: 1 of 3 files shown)


Juanlu001:
The result can be seen here:

http://juanlu001-rtree.readthedocs.io/en/latest/

Juanlu001:
For some reason, pytest is not installed in the Python 3.3 environment. What do you want me to do about it?

sgillies:
@Juanlu001 could you please add a `- pip install pytest` just before `- pip install -e .` in the .travis.yml file? I think that should do it.

Juanlu001:
The problem was with using `pytest` in Python 3.3, instead (I guess) `py.test`. I fixed it by using `python -m pytest`.

PR #110, opened 2018-09-18
Migrate to MIT license

### LINKED ISSUE #97: Migrate to MIT License?
> libspatialindex changed from a LGPL to a MIT license as of the 1.8.0 release. For most situations, this should have no impact on the library’s use, but it should open it up for usage in situations that otherwise might have been problematic. Versions of libspatialindex prior to 1.8.0 were licensed LGPL 2.0, with the license description on this file. The codebase has been been updated, with licensing information replaced in headers and source files, to use the MIT license as of the 1.8.0+ release.
> This change was made to support the inclusion of software depending on libspatialindex in static linking-only environments such as embedded systems and Apple’s iOS. libspatialindex versions prior to 1.8.0 will continue to live on as LGPL software, and developers can continue to contribute to them under terms of that license, but the main development effort, and ongoing maintenance, releases, and bug applications, will move forward using the new MIT license at http://github.com/libspatialindex/libspatialindex

[(source)](https://libspatialindex.github.io/#licensing-history)

yeesian:
closes #97


sgillies:
Since it's got your 👍 @hobu, I'm going to merge.

docs/source/changes.txt
```diff
@@ -3,6 +3,11 @@
 Changes
 ..............................................................................
 
+0.9: 2018-09-18
```
sgillies: @yeesian would you be willing to change the version to "Next" with no date? I think the release still needs to be decided.

PR #113, opened 2019-07-09
Adds Index_Contains_id, Index_Contains_obj methods

leobuskin:
Paired with PR https://github.com/libspatialindex/libspatialindex/pull/154


hobu:
This will need guards to prevent it from being used when the symbol does not exist in the library. Continuous integration currently fails due to this.

leobuskin:
@hobu what's the preferred way to wrap it with guards? I didn't use ctypes too much, to be frank, so not sure about best practices here.

hobu:
Use ctypes to check if the function exists in a try/except block and then add a boolean that protects whether or not to define the primitives.

hobu:
I shut off the doctests for now, but this is now conditionally available on libspatialindex 1.9.3 builds.

PR #117, opened 2019-10-29
Add initial support for TPR-Tree type

sdhiscocks:
Currently this handles the extra parameters by using a tuple of coordinates, velocities and times.

Before completing documentation and some tests, I wanted to check this approach is okay. Alternative would be to subclass `Index` to create specific `TPRtree` and `TPRContainer` classes.

TODO:
- [x] Documentation
- [x] Tests


hobu:
* you will need to adjust core.py to conditionally bring in symbols such as `Index_DeleteTPData`.
* I think it is fine to drop Python 2.x support. Update the travis configuration as necessary.
* Would it be possible to convert the test(s) in https://github.com/libspatialindex/libspatialindex/tree/master/test/tprtree to Python? It might be good verification that everything is making it through
* What about MVRTree?

sdhiscocks:
I've looked at converting the tests for TPR tree to Python, but seeing some issues. What seems to be happening is deletions from the tree are failing. In attempting to debug, it appears to only occur when leaf size is less than the data set size.

I'm wondering if there is some issue with deletion finding the correct node in the tree?

sdhiscocks:
Just to add, you'll see from the tests I've created, that I run with different leaf size and data set size to show this issue occurring. i.e. `100 leaf-dataset size 1000` fails, but `100 leaf-dataset size 100`, `1000 leaf-dataset size 100` and `1000 leaf-dataset size 1000` all pass.

hobu:
@normanb Do you happen to remember any deletion issues of the TPRTree variant and the C API? @sdhiscocks is trying to port the manual TPRTree tests to Python here.

normanb:
@hobu @sdhiscocks I don't remember any deletion issues, but there could well be with the way a TPR tree works. I used the TPR tree a lot in small demos but never in a production environment.

sdhiscocks:
So looks like the issue is with [`containsRegionAfterTime` returning false here](https://github.com/libspatialindex/libspatialindex/blob/1896a183afdff3f5845742ce3027553a88e13c3a/src/tprtree/Index.cc#L87) for all nodes.

To test, in _libspatialindex_ I created a `findLeafDel` method in `Index.cc` to be sure it's only called on deletion, and removed the if statement. Tests now pass, but obviously results in inefficient deletion.

sdhiscocks:
I've resolved the issue with the tests (it was generating additional delete operations).

hobu:
All the python versions less than 3.7 are failing on your test. Fix this up and it should be good to merge.

sdhiscocks:
Thanks @hobu. So I've fixed one issue with the tests, but other failures look to be present on _master_ branch. Adding a _skipif_ to these similar to the one I added to TPR tests should work. Happy to add a commit fixing that to this or a separate PR?

I'll just add a commit which adds some documentation first if that's okay? Hope to find some time tomorrow to do this.

hobu:
fixing it here is just fine.

I made a 0.9.0 milestone and I am hoping to cut the release by the end of the month. I have have success using `conda press` to build full wheels, and I will use that to attempt to do so for the Rtree release.

PR #129, opened 2019-12-09
Refactor tests to use Python unittest

hobu:


sgillies:
@hobu lots of .py tests deleted, is that intended?

hobu:
Yes, I collated all of the doctest bits into test_index.py. I'm trying to clean this up so it will be easier to add tests and maintain.

PR #136, opened 2019-12-20
Flake8 linting of core modules (`rtree/index.py` and `rtree/core.py`, ignoring E501)

sr-murthy:
Flake8 linting of core modules (`rtree/index.py` and `rtree/core.py`, ignoring E501)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,9 @@ addons:
       - libspatialindex-c3
 
 install:
+  - pip install flake8
   - pip install -e .
 
 script:
+  - flake8 --ignore=E501 --exclude=rtree/__init__.py rtree/
   - python -m pytest --doctest-modules rtree tests/test_*
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@ Rtree
 =====
 
 [![Build Status](https://travis-ci.org/Toblerity/rtree.svg)](https://travis-ci.org/Toblerity/rtree)
+[![PyPI version](https://badge.fury.io/py/Rtree.svg)](https://badge.fury.io/py/Rtree)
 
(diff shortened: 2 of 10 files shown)


hobu:
Add the linter to the CI testing?

sr-murthy:
This would be fine, except that the linting command does produce two undefined name errors in `rtree/index.py` that I'm currently ignoring

    $ flake8 --ignore=E501 rtree/index.py 
    rtree/index.py:15:13: F821 undefined name 'xrange'
    rtree/index.py:16:20: F821 undefined name 'basestring'

These two errors are a result of the conditional re-naming of `xrange` and `basestring` if the Python version is 2 starting here

https://github.com/Toblerity/rtree/blob/master/rtree/index.py#L14

This is perfectly fine functionally, but obviously it does generate the F821 errors if using Python >= 3. It should not generate those errors if using Py2. To make the linting step work for both Py2 and Py3 you could obviously wrap the original command in an appropriate conditional shell statement

hobu:
> These two errors are a result of the conditional re-naming of xrange and basestring if the Python version is 2 starting here

That's yuck. We can probably ditch the Python support anyway. Please make another PR that tosses that.

sr-murthy:
You would not want to exclude F821 entirely because it is useful to catch this generally.

sr-murthy:
OK, no problem.

sr-murthy:
I rebased this branch against the Py2 branch, and updated the CI scripts to include a step to `pip install` Flake8 and run it, before running the tests. It seems to be fine.

sr-murthy:
It may be better to create a separate step inside the CI scripts for Flake8 linting - ATM it is a substep inside the `pytest` step.

PR #137, opened 2019-12-20
Drop Python 2 support (changes in `rtree/index.py` module)

sr-murthy:
* Drop Python 2 support (changes in `rtree/index.py` module)

### DIFF
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -1507,7 +1507,7 @@ class Property(object):
     """Reinsert factor"""
 
     def get_filename(self):
-        return core.rt.IndexProperty_GetFileName(self.handle)
+        return core.rt.IndexProperty_GetFileName(self.handle).decode()
 
 
     def set_filename(self, value):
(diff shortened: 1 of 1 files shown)


hobu:
It looks like tests need to be updated. If there are any more Python 2-specific items, please remove them in this branch. It's time to let it go! ❄️

sr-murthy:
Yes, I've just updated three index test cases that needed to be fixed.

rtree/index.py
```diff
@@ -5,24 +5,12 @@
...
+import pickle
 
 import sys
-if sys.version_info[0] == 2:
-    range = xrange
-    string_types = basestring
-elif sys.version_info[0] == 3:
-    string_types = str
 
+string_types = str
```
mwtoews: Rather than define this here, on line 223 just replace `string_types` with `str`

sr-murthy: Makes sense. I've just updated the PR.

sr-murthy:
Closing as changes in this PR have now been incorporated into the Flake8 linting PR, which I've just merged.

PR #139, opened 2020-01-01
Add index size and custom `__repr__` methods for index classes

sr-murthy:
* Add `get_size` method for `rtree.index.Index` and `rtree.index.RtreeContainer`, which returns a count of the number of index objects/entries
* Add custom `__repr__` for both these classes that displays the current index bounds and the size count when the object is printed out to the console

### DIFF
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -147,7 +147,7 @@ class Index(object):
 
             >>> idx = index.Index(properties=p)
             >>> idx  # doctest: +ELLIPSIS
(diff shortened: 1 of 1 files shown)


sr-murthy:
@mwtoews Some CI checks appear to be flaky, e.g. Conda Linux Python3.6. I wonder if you have access to the CI system.

mwtoews:
it seems it just needed a re-run, checks are good now.

PR #148, opened 2020-03-03
Wheels can be created (for Linux)

Helveg:
* Only tested on my own Linux machine
* The ctypes loading is suboptimal but I have no experience with it.
* Don't know whether ALL of the libspatialindex install files are required, but currently they are all packaged along.
* I didn't make a fallback mechanism if the library is not in the `lib` dir, for example should someone use a different installation method than the wheels.

### DIFF
--- a/rtree/.gitignore
+++ b/rtree/.gitignore
@@ -1,0 +1,2 @@
+include
+lib
--- a/rtree/core.py
+++ b/rtree/core.py
@@ -81,7 +81,7 @@ if os.name == 'nt':
 
     def _load_library(dllname, loadfunction, dllpaths=('', )):
         """Load a DLL via ctypes load function. Return None on failure.
-        Try loading the DLL from the current package's directory first,
+        Try loading the DLL from the current package directory first,
         then from the Windows DLL search path.
         """
         try:
--- a/rtree/core.py
+++ b/rtree/core.py
@@ -81,11 +81,11 @@ if os.name == 'nt':
 
     def _load_library(dllname, loadfunction, dllpaths=('', )):
         """Load a DLL via ctypes load function. Return None on failure.
-        Try loading the DLL from the current package directory first,
+        Try loading the DLL from the current package's lib directory first,
         then from the Windows DLL search path.
         """
         try:
-            dllpaths = (os.path.abspath(os.path.dirname(__file__)),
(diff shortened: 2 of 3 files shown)


Helveg:
@hobu the Azure pipelines will have to be adapted so that they provide the required files in the `include` and `lib` folders of the `rtree` package directory.

hobu:
None of the linux or OSX code tests pass with these changes. We can't merge anything until we clear the issues.

Helveg:
@hobu, ofcourse they don't, they expect the `rtree/include` and `rtree/lib` folders to contain the necesary .so files. You have to adapt the Azure pipelines to copy the libspatialindex build artifacts there. Can you do this?

Helveg:
Another possibility is that we generate the wheels in all of these pipelines and after the unit testing we create a new virtual env and install and test the wheel we just generated, and unit test that again :)

hobu:
> Another possibility is that we generate the wheels in all of these pipelines and after the unit testing we create a new virtual env and install and test the wheel we just generated, and unit test that again :)

That would be completely reasonable. We could add another builder too. It all needs to be automated.

Helveg:
I noticed that the setup seems to break if the files are moved, and that in general it's very directory sensitive. See the `HACK` comment. Are these `.so` files not portable? If this approach is to work the directory they are deployed to is not allowed to matter, all the installed files will only be in the same relative position to eachother, but not in the same absolute position. 

I know for sure that .so files should technically be able to be moved around, because in another library (pytransform) I have done so. 

Anyways, if my suspicions are correct is there something about the build process of these so files that we can change to fix this?

hobu:
I have made a really hard go at this in #152 but https://github.com/pypa/auditwheel/issues/197#issue-500857416 is a blocker right now because of the way Rtree uses ctypes.

Helveg:
And is it required that Rtree uses ctypes in that specific way?

hobu:
When auditwheel and relocate can be made aware of additional shared libraries that are dlopen'd (like Rtree does to libspatialindex through ctypes), my configuration that I added to the Github Actions configuration could be resurrected.

Helveg:
I'm closing this PR as PR #156 can also create wheels for Linux, but over at that PR it's also integrated into the CI pipelines.

Issue #135, opened 2019-12-19
`undefined symbol: Error_GetLastErrorNum` when dependabot checks dependancies


bryanculbertson:
[Dependabot](https://dependabot.com/) stopped being able to check dependancies when rtree >0.9 is a requirement. According to their logs the failure is `undefined symbol: Error_GetLastErrorNum` when trying to pip install `rtree`.

I have filed an issue with Dependabot, but haven't heard back from them yet. Is there another fix for this other than getting Dependabot to install the latest `libspatialindex`?

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/tmp/tmpbs28h248build/rtree/setup.py", line 3, in <module>
    import rtree
  File "/tmp/tmpbs28h248build/rtree/rtree/__init__.py", line 1, in <module>
    from .index import Rtree
  File "/tmp/tmpbs28h248build/rtree/rtree/index.py", line 6, in <module>
    from . import core
  File "/tmp/tmpbs28h248build/rtree/rtree/core.py", line 145, in <module>
    rt.Error_GetLastErrorNum.restype = ctypes.c_int
  File "/usr/local/.pyenv/versions/3.7.5/lib/python3.7/ctypes/__init__.py", line 377, in __getattr__
    func = self.__getitem__(name)
  File "/usr/local/.pyenv/versions/3.7.5/lib/python3.7/ctypes/__init__.py", line 382, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
AttributeError: /usr/local/.pyenv/versions/3.7.5/bin/python3.7: undefined symbol: Error_GetLastErrorNum
```

hobu:
libspatialindex_c is not installed or findable.

hobu:
see https://github.com/Toblerity/rtree/issues/132

bryanculbertson:
Thanks @hobu. If you have some more time to help, I would like to improve my bug report to Dependabot.

Is the issue here that `libspatialindex` was not required to pip install `rtree` 0.8, but is required to pip install rtree 0.9?

Or is the issue that rtree 0.9 requires libspatialindex `1.9.3` and doesn't work with `1.8.5`?

hobu:
Rtree has always required libspatialindex. Rtree 0.9.3 will work with either libspatialindex 1.9.3 or 1.8.5. 

Rtree might be looking in a slightly different location or Dependabot is not specifying the need for libspatialindex if Rtree is selected. I'm not sure.

PR #149, opened 2020-03-06
Don't import rtree in setup.py

mikedh:
If you try to `pip install rtree` without `libspatialindex` the pip install fails, because `setup.py` imports `rtree`. If you didn't have this, `libspatialindex` could be installed later without killing the pip install. PR'ing as I ran into https://github.com/Toblerity/rtree/issues/135 as well and this prevents the error on install. 

Setup only needs rtree to set the version, this changes it to get that version directly from `__init__.py`. This also changes whitespace in `setup.py` to match pep8.

### DIFF
--- a/rtree/__init__.py
+++ b/rtree/__init__.py
@@ -1,5 +1,1 @@
 __version__ = '0.9.4'
-
-# from .index import Rtree
-
-# from .core import rt
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,17 @@
 #!/usr/bin/env python
-from setuptools import setup
-import rtree
 import os
 
+from setuptools import setup
 import itertools as it
 
 # Get text from README.txt
 with open('docs/source/README.txt', 'r') as fp:
     readme_text = fp.read()
+
+# Get __version without importing
+with open('rtree/__init__.py', 'r') as fp:
+    # get and exec just the line which looks like "__version__ = '0.9.4'"
+    exec(next(line for line in fp if '__version__' in line))
 
 extras_require = {
     'test': ['pytest>=3', 'pytest-cov', 'numpy']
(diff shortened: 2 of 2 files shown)


sja45uk:
My problem on Ubuntu 19.10 in a python3.7.5 environment was with:

pip install 'sharedstreets[dataframe]'

OR

pip install rtree
Collecting rtree
  Using cached Rtree-0.9.4.tar.gz (62 kB)
    ERROR: Command errored out with exit status 1:
. . .
File "/usr/lib/python3.7/ctypes/__init__.py", line 382, in __getitem__
        func = self._FuncPtr((name_or_ordinal, self))
    AttributeError: /home/ . . . /bin/python3.7: undefined symbol: Error_GetLastErrorNum

My workaround was:

sudo apt install libspatialindex-dev
pipenv install rtree
pip install 'sharedstreets[dataframe]'

craigds:
Any chance of a release for this?

This breaks dependabot in projects using rtree, and no workaround is possible since dependabot isn't very customisable (we can't install extra libs in the dependabot environment as far as I know)

hobu:
I was waiting on @mikedh to finish up the wheel-building activity before releasing.

craigds:
@hobu: hate to nag, but any idea of when that will happen and the release will be made?

hobu:
> any idea of when that will happen and the release will be made?

I was really hoping for the wheel issue to be knocked out. Most of the traffic on this repository is people wanting wheels, and my attempts at providing them were not successful. I invested quite a bit of time in developing Github Actions to allow someone who understands the wheel dark arts to have infrastructure to implement it in an automated fashion. 

If @mikedh doesn't have bandwidth for wheels in the next couple of weeks, I can snap off a release, but every time we do that we get a flood of "where's the wheels?" ticket traffic.

PR #156, opened 2020-05-25
Cmake install

Helveg:
* Switched setup.py to a cmake based install script
*  Updated ci scripts to build several different build artifacts (wheel, egg & source)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,7 @@ class cmake_build(build_ext):
 
 
 setup(
-    name='Rtree-linux',
+    name='Rtree',
     version=__version__,
     description='R-Tree spatial index for Python GIS',
     license='MIT',
--- a/setup.py
+++ b/setup.py
@@ -70,10 +70,15 @@ class cmake_build(build_ext):
         # ... setuptools is an enigma monkey patched on a mystery
         if not os.path.exists(dest_path):
             os.makedirs(dest_path, exist_ok=True)
+
+        release_path = os.path.join(source_path, "Release")
+        if os.path.exists(release_path):
+            source_path = release_path
         self.copy_tree(source_path, dest_path)
 
+
 setup(
-    name='Rtree',
+    name='Rtree-linux',
     version=__version__,
     description='R-Tree spatial index for Python GIS',
     license='MIT',
--- a/setup.py
+++ b/setup.py
@@ -49,9 +49,6 @@ class cmake_build(build_ext):
 
         build_args = ['--config', 'Release']
 
-        # Assuming Makefiles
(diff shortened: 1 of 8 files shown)


Helveg:
@hobu The builds don't seem to use the `ci/*` scripts at all? It seems that I have to modify the `.github` workflow instead, that's probably what you meant with the GitHub actions :'D

Helveg:
@hobu How could I inspect the generated artifacts of the builds? I'd like to check if they contain the `*.so` files. On all machines I've tested they do but the lib seems to be missing here.

Helveg:
Ok so after a tedious debugging process with a round-trip time of 2 minutes per explorative thought I found that when we test using pytest, we are in the root of our repository when we ask python to `import rtree`, because it finds a directory called `rtree` in there (namely the source directory) it opens that. The ctypes loading mechanism then doesn't find a `lib` folder, so tries to find the library elsewhere. It's not installed on the system --> The `Error_GetLastErrorNum` error occurs because the library isn't found.

But, what **should** happen is that we go look for the **installed** rtree package, which **does** contain a lib folder with the required libraries. A simple change of directories to a directory without an `rtree` folder as direct child should allow us to find the installed package and to test that instead.

Helveg:
I've opened libspatialindex/libspatialindex#190 to deal with the current MacOS build failure. To solve my own issue I'm releasing the wheels on PyPI.

A solution for this PR to pass the checks on Mac could be to install libspatialindex from the package as before and to release build artifacts for MacOS without libspatialindex included, also like before.

setup.py
```diff
@@ -70,10 +70,15 @@ def run(self):
...
+
+        release_path = os.path.join(source_path, "Release")
+        if os.path.exists(release_path):
+            source_path = release_path
         self.copy_tree(source_path, dest_path)
 
+
 setup(
-    name='Rtree',
+    name='Rtree-linux',
```
hobu, line 81: I don't understand renaming the library here.

hobu (changes requested):

setup.py
Helveg: Oh, I was working on my fork to release the 3.9 wheels for Windows & Linux! I didn't realise that branch was linked to this PR :) I'll revert this. I've recently asked someone to help me look into the Mac wheels which would let us finish this PR, so fingers crossed

hobu:
Closing for now. I think the new wheel infrastructure that @mikedh implemented accomplishes all that this effort tried to do. Please reopen if not.

Helveg:
Yup, I think so too :) We got pretty close, if I had a Mac device we'd probably have gotten there ^_^

PR #197, opened 2021-12-18
Hack to fix pickling

### LINKED ISSUE #87: Pickling and unpickling messes up Index
Pickling and unpickling renders the Index unusable. The test_pickle.py does not test the correct properties, should check Index.bounds as well. Code to reproduce:

```import cPickle as pickle
from rtree import index
import numpy as np

lonR = [-2.,9.]
latR = [50.,54.]
lonD = 0.125
latD = 0.125
lon = np.arange(lonR[0],lonR[1],lonD)
lat = np.arange(latR[0],latR[1],latD)
	
rtree = index.Index()
i = 0
for y in lat:
	for x in lon:
		rtree.insert(i, (x,y,x+lonD,y+latD))
		i += 1

unpickled = pickle.loads(pickle.dumps(rtree))
print(rtree.get_bounds())
print(unpickled.get_bounds())

### PR DESCRIPTION
**Pull request by adamjstewart**
This is a hack that allows `Index` objects to be pickled. It works by copying the entire index into memory, pickling it, then rebuilding the entire index during unpickling. I hope that there is a more efficient, less hacky way to do this, but until then this at least lets users pickle rtree objects.

Fixes #87

### DIFF
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -299,13 +299,18 @@ class Index(object):
                                                               self.get_size())
 
     def __getstate__(self):
-        state = self.__dict__.copy()
-        del state["handle"]
-        return state
+        attrs = self.__dict__.copy()
+        del attrs["handle"]
+        entries = self.intersection(self.bounds, objects=True)
+        entries = [(item.id, item.bbox, item.object) for item in entries]
+        return attrs, entries
 
     def __setstate__(self, state):
-        self.__dict__.update(state)
+        attrs, entries = state
+        self.__dict__.update(attrs)
         self.handle = IndexHandle(self.properties.handle)
+        for item in entries:
+            self.insert(*item)
 
     def dumps(self, obj):
         return pickle.dumps(obj)
--- a/tests/test_index.py
+++ b/tests/test_index.py
@@ -143,13 +143,22 @@ class IndexProperties(IndexTestCase):
 
 class TestPickling(unittest.TestCase):
 
+    @unittest.expectedFailure
     def test_index(self):
         idx = rtree.index.Index()
+        idx.insert(0, [0, 1, 2, 3], 4)
         unpickled = pickle.loads(pickle.dumps(idx))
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: rtree/index.py [hunk 1]
```diff
@@ -299,13 +299,18 @@ def __repr__(self):
                                                               self.get_size())
 
     def __getstate__(self):
-        state = self.__dict__.copy()
-        del state["handle"]
-        return state
+        attrs = self.__dict__.copy()
+        del attrs["handle"]
+        entries = self.intersection(self.bounds, objects=True)
+        entries = [(item.id, item.bbox, item.object) for item in entries]
```
adamjstewart, line 305: When `interleaved=True`, this seems to require `item.bbox`, but when `interleaved=False`, this seems to require `item.bounds`. I'm not even sure why there are two names for the same thing...

hobu: indeed that seems like some kind of bug due to refactorings. It should be tracked down all the way to figure out what's going on.

adamjstewart:
> Can you provide a test for this pickle support? I don't really want to merge it until it's tested.

This PR already contains testing for pickle support. We already had some testing, but we didn't actually test that the unpickled object matches the original object.

I don't actually want to merge this PR as is, as it's kind of hacky and I don't fully understand why this is necessary. I'm hoping that someone smarter than me can propose a better solution.

hobu:
It would be some work, but I think you have everything needed in the libspatialindex C API to make a custom storage type for Python. That could be wired into more common Python things like buffers and such.

adamjstewart:
I know very little about C/Cython so it might be better for someone else to do that. The goal of this PR was to push forward the discussion in #87 by proposing something so hacky that someone would investigate alternatives to avoid my current solution 😆

adamjstewart:
I'm wondering if we should revive this PR. Something hacky and slow that works is better than something that doesn't work. Being able to pickle an index is really important for any multiprocessing code on macOS/Windows.

codeananda:
I assume this also doesn't work for cPickle?

PR #207, opened 2022-02-04
Add __len__ method to Index

### LINKED ISSUE #194: Easy way to get number of entries in index
I would like to be able to easily query the number of entries in an index. I would expect `len(index)` to work. However, the `__len__` attribute isn't defined. I recently discovered `index.get_size()`, which gives me exactly what I want, although it isn't documented. Could we rename `get_size` to `__len__`, or have `__len__` call `get_size`? Alternatively, could we document `get_size`?

I'm happy to open a PR to do this, but it'll have to wait until next week.

adamjstewart:
This PR adds a `__len__` method to `Index`, defining the length of an index as the number of entries in that index. This is analogous to the definition of length for Python sets.

This PR also documents and tests the `get_size` method which provides identical functionality. This method has never been documented before. Should we deprecate or remove it in favor of `__len__`? It's definitely more Pythonic to rely on these kinds of magic methods instead of `get_*` and `set_*` methods.

Closes #194 

@hobu

### DIFF
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -289,10 +289,23 @@ class Index(object):
                     self.insert(*item)
 
     def get_size(self):
+        """The number of entries in the index.
+
+        :return: number of entries
+        :rtype: int
+        """
         try:
             return self.count(self.bounds)
         except core.RTreeError:
             return 0
+
+    def __len__(self):
+        """The number of entries in the index.
+
+        :return: number of entries
+        :rtype: int
+        """
+        return self.get_size()
 
     def __repr__(self):
         return 'rtree.index.Index(bounds={}, size={})'.format(self.bounds,
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -2,6 +2,7 @@ import ctypes
 import os
 import os.path
 import pprint
+import warnings
 
 from . import core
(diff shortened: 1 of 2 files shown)


hobu:
>  Should we deprecate or remove it in favor of `__len__`? 

Deprecate is fine. Rtree comes from a time before there were properties in Python 😄, which is why there's `get_*`-style stuff.

> It's definitely more Pythonic 

What was Pythonic 15 years ago is not necessarily the same thing as today. Rtree could definitely use a thorough once over in that regard.

EwoutH:
For some reason this warning is still triggered in rtree's [own CI](https://github.com/Toblerity/rtree/actions/runs/9876463917/job/27275625130?pr=322#step:6:23).
```
=============================== warnings summary ===============================
rtree/index.py::rtree.index.Index.__init__
  /home/runner/work/rtree/rtree/rtree/index.py:284: DeprecationWarning: index.get_size() is deprecated, use len(index) instead
    warnings.warn(
```
Is this expected? Could it be that a test is out of date?

adamjstewart:
Fixed by https://github.com/Toblerity/rtree/pull/323

PR #208, opened 2022-02-04
pytest: update deprecated fixture

adamjstewart:
Running the unit tests with pytest results in the following deprecation warnings:
```
tests/conftest.py:12
  /Users/Adam/Downloads/rtree/tests/conftest.py:12: PytestDeprecationWarning: @pytest.yield_fixture is deprecated.
  Use @pytest.fixture instead; they are the same.
    @pytest.yield_fixture(autouse=True)
```
This PR updates the fixture to its modern name.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,7 +9,7 @@ data_files = [
 ]
 
 
-@pytest.yield_fixture(autouse=True)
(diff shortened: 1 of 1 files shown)


adamjstewart:
@hobu this one should be good to go

PR #214, opened 2022-02-10
Bump required version of libspatialindex

### LINKED ISSUE #211: Minimum supported version of libspatialindex
According to:

* `DEPENDENCIES.txt`
* `docs/source/install.txt`
* `rtree/index.py`

the minimum supported version of lispatialindex is 1.7.0. However, according to:

* `docs/source/README.txt`

the minimum supported version of libspatialindex is 1.8.5.

Which is it? We should make the documentation uniform.

adamjstewart:
Closes #211

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,5 +3,5 @@ channels:
 - defaults
 - conda-forge
 dependencies:
-- python>=3.5
-- libspatialindex
+- python>=3.7
+- libspatialindex>=1.8.5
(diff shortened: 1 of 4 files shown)


environment.yml
```diff
@@ -3,5 +3,5 @@ channels:
 - defaults
 - conda-forge
 dependencies:
-- python>=3.5
-- libspatialindex
+- python>=3.7
```
adamjstewart, line 6: Missed this spot in #212

rtree/index.py
```diff
@@ -23,12 +23,11 @@
 
 __c_api_version__ = core.rt.SIDX_Version()
 
-major_version, minor_version, patch_version = [
```
adamjstewart, line 26: Tests are failing because I removed these attributes. Is this something you want to keep?

PR #215, opened 2022-02-11
Add partial type hints

### LINKED ISSUE #170: Type hinting for the code

adamjstewart:
I'm going to need a lot of help with this one. Many functions have no docstring so I have no clue what types they accept or return.

Related to #170


adamjstewart:
I have absolutely no idea how to type the ctypes stuff in `rtree/core.py`, might leave it untyped for now.

hobu:
> I have absolutely no idea how to type the ctypes stuff in rtree/core.py, might leave it untyped for now.

completely reasonable. I've thought about refactoring to use pybind11 and ditch the libspatialindex C API layer usage. This should make things perform better.

adamjstewart:
I think this is ready for review now. We're still a long way from `mypy --strict` passing, but many of the important user-facing functions now have type hints. For the remaining undocumented functions, we'll need someone who knows better what types they accept and return.

scripts/visualize.py
```diff
@@ -2,8 +2,8 @@
 
 import sys
 
-import ogr
 from liblas import file
```
hobu, line 5: This thing should really be dropped or updated to use laspy for fetching information. It is a very old holdover from a long time ago.

hobu (approved):

PR #217, opened 2022-02-14
Use separate GitHub Action step for linting

adamjstewart:
There's no need to run our linting tests on all possible combinations of OS/Python/libspatialindex versions. This PR adds a separate step solely for linting.

Questions:

* Do we want `collect-artifacts` to depend on `flake8`?
* I'm planning on adding black and isort in addition to flake8. Do we want a single `linting` step or separate steps for `flake8`, `black`, and `isort`? In TorchGeo we actually have separate files for linting and testing: https://github.com/microsoft/torchgeo/tree/main/.github/workflows


setup.cfg
```diff
@@ -34,3 +34,6 @@ packages = rtree
 
 [options.package_data]
 rtree = lib
+
+[flake8]
+max-line-length = 88
```
adamjstewart, line 39: Black uses a line length of 88 by default. We could instead change black's line length to 79/80 to match flake8, but I think most projects are simply using what black does these days.

hobu: I've no strong opinion. The purpose of the linting is to try to prevent the codebase's style from diverging from norms.

PR #222, opened 2022-02-21
Fix C API mismatches vs. libspatialindex

### LINKED ISSUE #220: Apparent C API mismatch causes test failure on big-endian architecture (s390x)
This comes from a [downstream Fedora Linux bug report](https://bugzilla.redhat.com/show_bug.cgi?id=2055249), in which `tests/test_index.py::IndexSerialization::test_interleaving` and `tests/test_index.py::IndexStream::test_stream_input` were failing on `s390x` (Fedora’s only big-endian primary architecture). This wasn’t noticed for a long time because the `python-Rtree` package is `noarch` and is normally built once on any available builder rather than once per architecture.

The submitter [concluded](https://bugzilla.redhat.com/show_bug.cgi?id=2055249#c1):

> I think I found the bug here - the size_t is getting chopped in half on 64-bit platforms, and that matters more for big endian machines.
> 
> I think the prototypes are wrong for size_t values in other signatures, but this seems to be the only one causing test failures. Maybe it only matters if it's a size_t pointer - I'm not sure.

They also pointed out that the change was intentional, https://github.com/Toblerity/rtree/pull/77, and claimed to *fix* a mismatch—but the [callback signature in question](https://github.com/libspatialindex/libspatialindex/blob/8ee223632f95c81f49f5eb2d547ad973475c4601/include/spatialindex/capi/sidx_api.h#L42-L44) appears to have always had a `size_t *` as its last parameter in `libspatialindex`.

The following change fixed the test failure:

```diff
diff -Naur Rtree-0.9.7-original/rtree/core.py Rtree-0.9.7/rtree/core.py
--- Rtree-0.9.7-original/rtree/core.py	2020-12-24 07:38:19.000000000 -0800
+++ Rtree-0.9.7/rtree/core.py	2022-02-16 16:11:36.115401345 -0800
@@ -99,7 +99,7 @@
                             ctypes.POINTER(ctypes.POINTER(ctypes.c_double)),
                             ctypes.POINTER(ctypes.c_uint32),
                             ctypes.POINTER(ctypes.POINTER(ctypes.c_ubyte)),
-                            ctypes.POINTER(ctypes.c_uint32))
+                            ctypes.POINTER(ctypes.c_size_t))
 
 rt.Index_CreateWithStream.argtypes = [ctypes.c_void_p, NEXTFUNC]
 rt.Index_CreateWithStream.restype = ctypes.c_void_p
```

However, the submitter noted that there seemed to be other prototype mismatches with `libspatialindex`’s C API. I audited the wrapped functions and will report my findings in comments on this issue.

musicinmybrain:
This fixes #220 by correcting the mismatches with the `libspatialindex` API that were noted in that issue and its comments.

### DIFF
--- a/rtree/core.py
+++ b/rtree/core.py
@@ -99,7 +99,7 @@ NEXTFUNC = ctypes.CFUNCTYPE(
     ctypes.POINTER(ctypes.POINTER(ctypes.c_double)),
     ctypes.POINTER(ctypes.c_uint32),
     ctypes.POINTER(ctypes.POINTER(ctypes.c_ubyte)),
-    ctypes.POINTER(ctypes.c_uint32),
+    ctypes.POINTER(ctypes.c_size_t),
 )
 
 rt.Index_CreateWithStream.argtypes = [ctypes.c_void_p, NEXTFUNC]
--- a/rtree/core.py
+++ b/rtree/core.py
@@ -73,12 +73,14 @@ def free_error_msg_ptr(result, func, cargs):
 # load the shared library by looking in likely places
 rt = finder.load()
 
+rt.Error_GetLastErrorNum.argtypes = []
 rt.Error_GetLastErrorNum.restype = ctypes.c_int
 
 rt.Error_GetLastErrorMsg.argtypes = []
 rt.Error_GetLastErrorMsg.restype = ctypes.POINTER(ctypes.c_char)
 rt.Error_GetLastErrorMsg.errcheck = free_error_msg_ptr  # type: ignore
 
+rt.Error_GetLastErrorMethod.argtypes = []
 rt.Error_GetLastErrorMethod.restype = ctypes.POINTER(ctypes.c_char)
 rt.Error_GetLastErrorMethod.errcheck = free_returned_char_p  # type: ignore
 
--- a/rtree/core.py
+++ b/rtree/core.py
@@ -323,7 +323,7 @@ rt.IndexProperty_SetDimension.restype = ctypes.c_int
 rt.IndexProperty_SetDimension.errcheck = check_return  # type: ignore
 
 rt.IndexProperty_GetDimension.argtypes = [ctypes.c_void_p]
-rt.IndexProperty_GetDimension.restype = ctypes.c_int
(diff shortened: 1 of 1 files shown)


hobu:
Tests aren't happy here with these changes quite yet.

musicinmybrain:
> Tests aren't happy here with these changes quite yet.

I noticed that, but haven’t had a chance to look into it yet. Since it Works on My Machine™ with spatialindex 1.9.3, I’ll have to take a little time to get the bottom of it. Strangely, the two failing tests are the same ones that prompted https://github.com/Toblerity/rtree/issues/220.

musicinmybrain:
I went ahead and pushed a commit that *should* fix this; see https://github.com/Toblerity/rtree/issues/220#issuecomment-1047827426. I’d rather not have to set up a local environment with libspatialite 1.8.5; if the tests pass this time, then I can avoid that. That said, I’m very much open to alternative approaches.

hobu:
Looks like a few lint things still, but this is converging

musicinmybrain:
Force-pushed with the most recent commit “blackened”.

I haven’t used mypy yet, but it looks like the initial assignment

```python
_nDataLength_type = ctypes.POINTER(ctypes.c_size_t)
```

locks in the type, and then conditionally reassigning

```python
_nDataLength_type = ctypes.POINTER(ctypes.c_uint32)
```

is “wrong.” I… didn’t expect it would work that way. I can rearrange the logic to make `mypy` happy.

musicinmybrain:
Force-pushed again; `mypy` is mollified.

hobu (approved):

hobu:
Is this ready to merge once it is ✅ ?

adamjstewart:
> I… didn’t expect it would work that way.

Yep, this is correct. Python is dynamically typed, but mypy can only check the type for statically typed objects. So changing the type of a variable is considered "bad" by mypy, but you can always specify the type as `Union[c_size_t, c_uint32]` to note that is can either be one or the other, or `typing.Any` to denote that it is dynamically typed and the type checker should ignore it.

musicinmybrain:
> Is this ready to merge once it is white_check_mark ?

As far as I can tell, yes.

PR #225, opened 2022-02-22
Apply pyupgrade --py37-plus --keep-runtime-typing, with mods

mwtoews:
Modernize the code base to Python 3.7+ using [pyupgrade](https://github.com/asottile/pyupgrade) with a few manual modifications for string formatting. Also change a message with "LASError" (from libLAS) to simply "Error".

The initial command was:
```
pyupgrade --py37-plus --keep-runtime-typing `find . -name "*.py"`
```

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 #
 # Rtree documentation build configuration file, created by
 # sphinx-quickstart on Tue Aug 18 13:21:07 2009.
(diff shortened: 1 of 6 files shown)


adamjstewart (approved):
Never heard of this tool before, this is great! This would have really helped with one of my recent projects.

adamjstewart:
There's also some `PY3` code in `tests/test_index.py` that's worth removing. I can do it in a separate PR unless you want to do it in this PR.

hobu:
doesn't matter to me

hobu:
merged it. Make a new PR for the test tweak if you want.

Issue #191, opened 2021-07-19
OSError: OSError: could not find or load spatialindex_c-64.dll


msaifp:
File "CSV_to_Shapefile_Converter.py", line 11, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "PyInstaller\loader\pyimod03_importers.py", line 546, in exec_module
  File "geopandas\__init__.py", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "PyInstaller\loader\pyimod03_importers.py", line 546, in exec_module
  File "geopandas\_config.py", line 126, in <module>
  File "geopandas\_config.py", line 112, in _default_use_pygeos
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "PyInstaller\loader\pyimod03_importers.py", line 546, in exec_module
  File "geopandas\_compat.py", line 202, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "PyInstaller\loader\pyimod03_importers.py", line 546, in exec_module
  File "rtree\__init__.py", line 9, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "PyInstaller\loader\pyimod03_importers.py", line 546, in exec_module
  File "rtree\index.py", line 6, in <module>
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
[...]

PR #236, opened 2022-03-31
Fixing could not find or load spatialindex_c-64.dll with conda (#191)

okhoma:
In the newer version of conda there is no longer `"conda" in sys.version`, unless you install python with -c conda-forge.

### DIFF
--- a/rtree/finder.py
+++ b/rtree/finder.py
@@ -37,8 +37,7 @@ def load() -> ctypes.CDLL:
         lib_name = f"spatialindex_c-{arch}.dll"
 
         # add search paths for conda installs
-        if "conda" in sys.version:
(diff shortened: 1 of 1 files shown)


hobu:
That's frustrating. Is there a more standard way of determining if we're are running under conda and its non-standard (or at least different-standard) system layout?

okhoma:
One could check environment variable CONDA_PREFIX that should only be defined if running in activated conda environment. It should be the same as sys.prefix.

okhoma:
Why do we absolutely need to check if running under conda environment before checking that folder? If it does not exist, it will be just skipped.

hobu:
Addressed in #239

PR #240, opened 2022-04-04
get ready for 1.0.0

hobu:
This is the final release prep @adamjstewart


adamjstewart:
Any comments on the decision to jump to 1.0.0? Will we be using semver from now on?

hobu:
The milestone had been 1.0.0 for almost two years. The plan had always been to make the next release be 1.0.0.

PR #243, opened 2022-04-18
Make Rtree PEP 561-compliant, so that Mypy finds type hints

oderby:
After #215, I was excited to take advantage of type hints in my projects that
use Rtree. However, after upgrading to v1.0.0, mypy was giving [import
errors]. According to those docs, mypy will only infer/lookup types from
installed packages if they are [PEP 561] compliant.

Fortunately, all that seems to [involve] is the inclusion of a `py.typed` marker
file in the installed package. This PR is my attempt at doing just that. In my
local testing, this seems to resolve the above Mypy import warnings.

I'm new to setuptools and distributing sdist wheels though, so I'm not sure how
to confirm this file will be included in the built wheels? From reading the
[setuptools docs], I think this is right, but probably worth review from someone
who knows these things better.

[import errors]: https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
[PEP 561]: https://peps.python.org/pep-0561/
[involve]: https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages
[setuptools docs]: https://setuptools.pypa.io/en/latest/userguide/datafiles.html

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ include_package_date = True
 packages = rtree
 
 [options.package_data]
-rtree = lib, py.typed
+rtree = lib
 
 [flake8]
 max-line-length = 88
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ include_package_date = True
 packages = rtree
 
 [options.package_data]
-rtree = lib
+rtree = lib, py.typed
 
 [flake8]
 max-line-length = 88
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include *.md *.txt
 include MANIFEST.in
+include rtree/py.typed
 recursive-include tests *
 recursive-include docs *
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,4 @@
 include *.md *.txt
 include MANIFEST.in
-include rtree/py.typed
 recursive-include tests *
 recursive-include docs *


setup.cfg
```diff
@@ -36,7 +36,7 @@ include_package_date = True
 packages = rtree
 
 [options.package_data]
-rtree = lib
+rtree = lib, py.typed
```
adamjstewart, line 39: It looks like the modifications to `setup.cfg` aren't needed. You can confirm this by running `python -m build` (assuming that build is installed) and checking `Rtree.egg-info/SOURCES.txt` to see what gets included in the wheel.

adamjstewart (approved):
Thanks for catching this! I think you can remove the changes to `setup.cfg`, but otherwise this looks good. Guess I have a few other libraries where I need to do this...

setup.cfg
oderby: Sounds good - removed.

adamjstewart:
Actually, I was wrong. `MANIFEST.in` and `SOURCES.txt` only correspond to what is put in the sdist, it looks like `package_data` is what controls what is put in the wheel. Actually, if I remove the changes to `MANIFEST.in` and only include the changes to `setup.cfg`, it looks like `py.typed` ends up in both the sdist and wheel. You can check this by extracting the tarball and wheel in `dist/` after running `python -m build`.

Can you undo the last commit and remove the changes to `MANIFEST.in` instead? I would also be fine with keeping the changes to both, this stuff is more confusing than it should be.

adamjstewart (approved):

oderby:
Looks like the `Build / Build wheel on aarch64 for cp39-* (pull_request)` CI job timed out? Is there a way to re-run it?

PR #250, opened 2022-10-10
CI: Update used actions, add weekly schedule

EwoutH:
A bit of CI maintenance:
- Updates all the used actions (checkout, setup-python, cibuildwheel) to their latest versions.
- Clean up the Ubuntu build matrix and add multiple Python versions
- Adds a weekly scheduled run to all CI workflows, which run each Monday at 06:00 UTC. A periodically scheduled run can detect errors and warnings appearing in changes in upstream dependencies or the build environment. By having a scheduled run they can be traced back easier to a dependency or environment change, then if they would pop up in a random PR.
- Update to libspatialindex-c6 on Ubuntu to support the 20.04 and 22.04 LTS versions

The updated cibuildwheel version will also build Python 3.11 wheels for the next release.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,13 +6,17 @@ on:
   release:
     types:
       - published
+  workflow_dispatch:
+  schedule:
+    - cron: '0 6 * * 1'
+
 jobs:
   black:
     name: black
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
       with:
         python-version: '3.10'
     - name: Setup
@@ -24,8 +28,8 @@ jobs:
     name: flake8
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
       with:
         python-version: '3.10'
     - name: Setup
@@ -37,8 +41,8 @@ jobs:
     name: isort
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
       with:
         python-version: '3.10'
     - name: Setup
@@ -50,8 +54,8 @@ jobs:
     name: mypy
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-python@v2
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
       with:
         python-version: '3.10'
     - name: Setup
@@ -71,7 +75,7 @@ jobs:
         sidx-version: ['1.8.5','1.9.3']
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - uses: conda-incubator/setup-miniconda@v2
       with:
         channels: conda-forge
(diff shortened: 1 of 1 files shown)


EwoutH:
Especially if you leave it alone for months is very useful, since when you come back you can directly see if the CI is broken and if so when. The when tells you the difference between packages or environments updated that week, which helps with quickly pinpointing what broke the build.

It's like bisection but you don't have to do it yourself :)

EwoutH:
There was one small thing I was hoping for: With the update to [cibuildwheel](https://github.com/pypa/cibuildwheel), Python 3.11 wheels are now build by default! And the wheels build successfully, so that means rtree already supports Python 3.11.

Python 3.11 is now in release candidate phase, which means that wheels can be uploaded to PyPI (and is encouraged). NumPy, SciPy and Pandas already have Python 3.11 wheels on PyPI, among others.

That means the only thing that needs to be done to get Python 3.11 wheels on PyPI is merging this PR and tagging a new release!

Since many geospatial packages use rtree, this is really awesome, since they now can start working on Python 3.11 support before it's released!

EwoutH:
Dependabot is great! There are many other this that can change in a CI environment however. For example:
 - The default version of a package (think compilers and such) on the [GitHub Actions images](https://github.com/actions/runner-images) can change
 - The image used can change (`ubuntu-latest` is now 20.04, but will soon be 22.04)
 - A Python patch version can change
 - A used action can have minor updates
 - Dependencies that are not strictly defined can be updated

So having CI somewhat periodically running helps errors from piling up between runs.

What do you think about the Python 3.11 wheels?

adamjstewart:
> The default version of a package (think compilers and such) on the [GitHub Actions images](https://github.com/actions/runner-images) can change

How often does this happen? I would have figured compiler versions are locked to OS versions.

> The image used can change (ubuntu-latest is now 20.04, but will soon be 22.04)

We should probably use `ubuntu-20.04` instead of `ubuntu-latest` if we care about CI stability.

> A Python patch version can change

I recently hit this with mypy and the latest Python patch version which broke numpy support. So I know what you mean.

> A used action can have minor updates

This is why the GitHub Actions [workflow syntax docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses) recommend using commit hashes or full version identifiers. This becomes annoying to manually update unless you use dependabot to track it though.

> Dependencies that are not strictly defined can be updated

All dependencies should be strictly defined. Are we missing anything?

To be clear, I'm not opposed to anything in this PR, just saying that dependabot can also be used to achieve many of these same goals.

> What do you think about the Python 3.11 wheels?

Sounds great! Exciting to see rtree getting ahead of the curve.

EwoutH:
> How often does this happen? I would have figured compiler versions are locked to OS versions.

About once or twice a year for major versions.

> We should probably use `ubuntu-20.04` instead of `ubuntu-latest` if we care about CI stability.

I think using the latest is fine, one of the purposes of CI is to check if things work well in new/updated/different environments.

> I recently hit this with mypy and the latest Python patch version which broke numpy support. So I know what you mean.

> This is why the GitHub Actions [workflow syntax docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses) recommend using commit hashes or full version identifiers. This becomes annoying to manually update unless you use dependabot to track it though.

Personally I allow things to break time to time, that way you get better insight in what are weaknesses of your software. If a certain part needs fixing often, there might be something else going on (the design for example).

> All dependencies should be strictly defined. Are we missing anything?

Depending on your perspective, objectives and procedures for CI :).

> Sounds great! Exciting to see rtree getting ahead of the curve.

Great to hear! In that case I say let's merge and tag a v1.0.1 or v1.1.0 release to get those Python 3.11 wheels going!

adamjstewart (approved):

adamjstewart:
I'm fine with this, although probably don't have cycles to watch this. I'd also be willing to set up dependabot (in addition to this PR), but probably don't have cycles to watch that either.

adamjstewart:
I'll let @hobu decide, I'm a relatively new maintainer and don't have access to PyPI so can't make releases anyway.

hobu:
1.0.1 released. https://pypi.org/project/Rtree/

adamjstewart:
New type hints are already catching problems in my code: https://github.com/microsoft/torchgeo/actions/runs/3230077395/jobs/5288108730

hobu:
> New type hints are already catching problems in my code

You're welcome? 😉

hobu:
Of course the first week this build fires it fails due to arm64 timeouts. If it fails like that it is going to get ignored anyway.

PR #251, opened 2022-11-02
Add Dependabot configuration for GitHub Actions updates

EwoutH:
Add a Dependabot configuration that checks once a week if the GitHub Actions are still using the latest version. If not, it opens a PR to update them.

It will actually open very few PRs, since we only have major versions specified (like v3), so only on a major v4 release it will update and open a PR.

See [Keeping your actions up to date with Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"


adamjstewart (approved):
We use the same thing in [TorchGeo](https://github.com/microsoft/torchgeo/blob/main/.github/dependabot.yml) and it works well for us. For maximum stability we include the full major.minor.patch version.

hobu:
I would like to note that https://github.com/Toblerity/rtree/pull/250 has failed with timeouts at every weekly fire, and I have been appropriately ignoring it 😄 

It is fine to merge something like this, but I'm not signing up to instantly react to any failure of it. If other folks value that, they need to react with PRs to address things it finds.

EwoutH:
> It is fine to merge something like this, but I'm not signing up to instantly react to any failure of it.

That's perfectly fine! It's meant to make life easier, not harder, so use it however you see fit.

With the weekly run, you can now look back when it first starts failing, without too many changes having taken place between the previous build that did pass.

PR #253, opened 2022-11-03
CI: Simplify wheel building

EwoutH:
Refactor and simplify the wheel building. A single job now builds all `aarch64` wheels, using the [cibuildwheel](https://github.com/pypa/cibuildwheel) GitHub Action (following [these docs](https://cibuildwheel.readthedocs.io/en/stable/faq/#emulation)).

The regular wheels are also build by the GitHub Action.

Another advantage is that Dependabot will update the GitHub Action once a new cibuildwheel version is released.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -135,7 +135,7 @@ jobs:
       CIBW_REPAIR_WHEEL_COMMAND_MACOS: ""
       MACOSX_DEPLOYMENT_TARGET: 10.9
       CIBW_BUILD_VERBOSITY_MACOS: 3
-      CIBW_TEST_SKIP: "*-macosx_arm64 *-macosx_universal2:arm64 *-musllinux*"
+      CIBW_TEST_SKIP: "*-macosx_arm64 *-macosx_universal2:arm64 *-musllinux* *i686"
       CIBW_ARCHS_MACOS: "x86_64 arm64"
       CIBW_REPAIR_WHEEL_COMMAND_WINDOWS: ""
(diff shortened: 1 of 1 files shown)


EwoutH:
I noticed that the `install_libspatialindex.bash` script took a huge amount of time, and was run before each wheel build. For 10 wheels, that was over an hour alone.

However, the `install_libspatialindex.bash` script is independent of the Python version used to build the wheel. So is only needs to be run once, to work for all Python versions. Therefore I changed the `CIBW_BEFORE_BUILD_LINUX` environment variable to `CIBW_BEFORE_ALL_LINUX` to only run the `install_libspatialindex.bash` script once.

EwoutH:
@hobu, if you don't want to wait on the CI to pass, on my fork it runs perfectly time and builds wheels _a lot_ faster. See https://github.com/EwoutH/rtree/actions/runs/3385946454

hobu:
awesome. Thanks so much!

EwoutH:
I noticed some of the wheel testing jobs take very long, because NumPy is build from scratch, which takes very long. NumPy doesn't provide wheels for musllinux or Aarch64 Python 3.7, so on those build I disabled testing.

EwoutH:
Okay that last commit helped _a lot_! For the regular manylinux job, we're down to [9 minutes](https://github.com/EwoutH/rtree/actions/runs/3386622199/jobs/5626267093) instead of 1h 45m to 2h 15m.

EwoutH:
[It passes](https://github.com/EwoutH/rtree/actions/runs/3386622199) on my fork! The Linux aarch64 wheels use one job of about 30 minutes, instead of 5 jobs of 2+ hours.

So it's ready to be merged.

PR #254, opened 2022-11-03
CI: Only run on pushes to the master branch or CI change

EwoutH:
Only run the CI on pushes to the master branch when the CI configuration is changed. This saves some time, since our CI matrix is huge.

Still also runs on all PRs, releases and once a week.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,10 @@ name: Build
 
 on:
   push:
+    branches:
+      - master
+    paths:
+      - '.github/workflows/build.yml'
   pull_request:
   release:
     types:


EwoutH:
@hobu Please cancel this CI run, the configuration works, so we don't have to wait for it to finish.

hobu:
maybe it would be useful to push a cancel strategy ala https://github.com/PDAL/PDAL/blob/master/.github/workflows/conda.yml#L11 too

EwoutH:
Could be interesting indeed! Sometimes is nice to see exactly what changes between two commits CI wise, and rtree isn't in such active development that is matters that much, #254 combined with #253 should help a lot.

But if after those changes the CI is still a major bottleneck, that would be the next place to look. Alongside with splitting the wheel building into a separate workflow that runs less often.

PR #269, opened 2023-07-20
Move project metadata to pyproject.toml

mwtoews:
This PR moves the core project metadata to pyproject.toml ([PEP 621](https://peps.python.org/pep-0621/)) as supported by [setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html).

This is a close 1:1 conversion with a few minor differences:

- Remove C/C++ classifiers, since this repo doesn't have any
- Add Python 3.11 classifier (note that the CI infrastructure has not been updated, perhaps it should)
- Use project.url table to show links for Documentation and Repository

The package name is repeated in `setup.py` since there are some legacy tools that assume the package name is there. For example the [dependency graph](https://github.com/Toblerity/rtree/network/dependents) is currently empty, but should be resolved.

Also setup.cfg is removed, moving flake8 configuration to `.flake8`.


mwtoews:
The [dependency graph](https://github.com/Toblerity/rtree/network/dependents) is now resolved, listing a wide range of other projects.

adamjstewart:
Confirmed all differences to the generated wheels, looks good!

PR #271, opened 2023-07-21
Wheel is not a required setup dependency

adamjstewart:
Once upon a time, the Python Packaging Tutorial suggested to include wheel as a required dependency for all projects. However, wheel is only needed to build wheels, and https://github.com/pypa/packaging.python.org/pull/1050 removed this suggestion. I think we should just remove this.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=61", "wheel"]
+requires = ["setuptools>=61"]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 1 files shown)


adamjstewart:
Ah, well this is annoying. Wheel is needed at run-time for pip for certain actions (including `pip install` from source) but isn't listed as a dep (because it isn't needed to install wheels). Let's just leave the dep there for now.

mwtoews:
I was about to remove this too, but it is used in setup.py

PR #273, opened 2023-07-23
Drop Python 3.7, upgrade typing for 3.8+

### LINKED ISSUE #228: Comparison in test_index.py::IndexSerialization::test_interleaving sometimes fails
It seems that `tests/test_index.py::IndexSerialization::test_interleaving` sometimes [fails](https://github.com/Toblerity/rtree/runs/5326194380?check_suite_focus=true#step:8:3645):
```
  >       assert all(
              all(np.allclose(a, b) for a, b in zip(L, E))
              for L, E in zip(leaves, expected)
          )
...
  x = array([ 2, 92, 51, 55, 26, 95,  7, 81, 38, 22, 58, 89, 91, 83, 98, 37, 70,
         31, 49, 34, 11,  6, 13,  3, 23, 57, ... 54, 88, 72, 97, 60, 62, 48, 19,
         25, 76, 59, 66, 64, 79, 94, 40, 32, 46, 47, 15, 68, 10,  0, 80, 56,
         50])
  y = array([ 2., 92., 51., 55., 26., 95.,  7., 81., 38., 22., 58., 89., 91.,
         83., 98., 37., 70., 31., 49., 34., 11.,...48., 19., 25.,
         76., 59., 66., 64., 79., 94., 40., 32., 46., 47., 15., 68., 10.,
          0., 80., 56., 50., 30.])
  Error: Command ['sh', '-c', 'pytest -v /project/tests'] failed with code 1. 
  atol = 1e-08, rtol = 1e-05
  
  
      def within_tol(x, y, atol, rtol):
          with errstate(invalid='ignore'):
  >           return less_equal(abs(x-y), atol + rtol * abs(y))
  E           ValueError: operands could not be broadcast together with shapes (69,) (70,)
```
looking at the test data, it should have size 70, so for some reason `leaves` misses the last element `30`. Curious to see if this this issue is valid or intermittent. I can't replicate it locally.

_Originally posted by @mwtoews in https://github.com/Toblerity/rtree/issues/227#issuecomment-1050369770_

mwtoews:
This PR drops Python 3.7 which reached it's end-of-life in June 2023 ([PEP 537](https://peps.python.org/pep-0537/)).

Testing and some code is modified to assume Python 3.8+. Some of this was automated using `pyupgrade --py38-plus`.

Closes #228

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
       fail-fast: true
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
         sidx-version: ['1.8.5','1.9.3']
 
     steps:
--- a/DEPENDENCIES.txt
+++ b/DEPENDENCIES.txt
@@ -1,4 +1,4 @@
-- python 3.7+
+- python 3.8+
 - setuptools
 - libspatialindex C library 1.8.5+: 
   https://libspatialindex.org/
--- a/environment.yml
+++ b/environment.yml
@@ -3,5 +3,5 @@ channels:
 - defaults
 - conda-forge
 dependencies:
-- python>=3.7
+- python>=3.8
 - libspatialindex>=1.8.5
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ maintainers = [
 ]
 description = "R-Tree spatial index for Python GIS"
 readme = "README.md"
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 keywords = ["gis", "spatial", "index", "r-tree"]
 license = {text = "MIT"}
 classifiers = [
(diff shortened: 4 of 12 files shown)


adamjstewart (approved):
We could add `from __future__ import annotations` to all files, or remove it from all files, if we want to be more consistent.

mwtoews:
I'm unsure about that statement (whether to add or remove from all). I can go either way, depending on advice from typing pros (not me!)

adamjstewart:
I don't have any strong opinions. I used it at one point in TorchGeo then later decided to undo it. With pyupgrade it's so easy to change that it doesn't really matter anyway.

adamjstewart:
So the only purpose that `from __future__ import annotations` served in that file was to allow references to classes inside of class definitions. If you convert those to strings (i.e., `Item` -> `"Item"`) then that will make things work again. You'll also need to undo some of the pyupgrade changes, `|` isn't supported for type hints until Python 3.10.

mwtoews:
I'm rolling-back fe8437ed4f98147046f3a4108ad236d7cb91af60 and doing `git ls-files -z '*.py' | xargs -0 isort -a "from __future__ import annotations"`

PR #276, opened 2023-08-18
Refactor cibuildwheel setup

### LINKED ISSUE #270: Linux binary wheels have two copies of libspatialindex
Looking at [PyPI files for 1.0.1](https://pypi.org/project/Rtree/1.0.1/#files) it seems that the linux versions have two copies of libspatialindex.

For instance `Rtree-1.0.1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl` has:
```
.
├── rtree
│   ├── core.py
│   ├── exceptions.py
│   ├── finder.py
│   ├── index.py
│   ├── __init__.py
│   ├── lib
│   │   ├── libspatialindex_c.so
│   │   └── libspatialindex.so
│   └── py.typed
├── Rtree-1.0.1.dist-info
│   ├── LICENSE.txt
│   ├── METADATA
│   ├── RECORD
│   ├── top_level.txt
│   └── WHEEL
└── Rtree.libs
    └── libspatialindex-91fc2909.so.6.1.1
```
both Windows and macOS binary wheels only have one copy in `rtree/lib`.

On Linux, cibuildwheel uses [auditwheel](https://github.com/pypa/auditwheel) to copy external shared libraries. The cibuildwheel stuff should be fixed to not install `rtree/lib` for Linux, which may require changes to `rtree.finder`. Or, perhaps the bundled libraries in `rtree/lib` are preferred and auditwheel's `Rtree.libs` should be removed somehow.

Presently only the `rtree/lib` libraries are used from these PyPI binary wheels:
```
$ python -c "from rtree import finder; print(finder.load())"
<CDLL '/tmp/py39/lib/python3.9/site-packages/rtree/lib/libspatialindex_c.so', handle 56400f603870 at 0x7ff4184bb3a0>
```

mwtoews:
There are a few aims in this PR:

- Move cibuildwheel data from GitHub Actions variables to pyproject.toml
- Build each platform, but allow to work with any Python 3 version, since this is a Pure Python project
- Use tox to run pytest for all Python versions
- For Linux, remove duplicated `rtree/lib`; close #270
- Modify finder to look for auditwheel-built shared library (using `importlib.metadata.files`)

The repair_wheel.py script was modified from [cmake-python-distributions](https://github.com/scikit-build/cmake-python-distributions/blob/master/scripts/repair_wheel.py). Admittedly, there are several things that I don't understand, but have left alone. It uses `wheel unpack` and `wheel pack` to modify the binary wheels.

### DIFF
--- a/scripts/repair_wheel.py
+++ b/scripts/repair_wheel.py
@@ -64,7 +64,8 @@ def main():
             shutil.copyfile(file, tmpdir / file.name)
         (file,) = tmpdir.glob("*.whl")
 
-        # we need to handle macOS universal2 & arm64 here for now, let's use platform_tag_args for this.
+        # we need to handle macOS universal2 & arm64 here for now,
+        # let's use platform_tag_args for this.
         platform_tag_args = []
         if os_ == "macos":
             additional_platforms = []
@@ -85,7 +86,8 @@ def main():
             additional_platforms.append("macosx_{}_{}_arm64".format(*arm64_target))
 
             if target < (11, 0):
-                # They're were also issues with pip not picking up some universal2 wheels, tag twice
+                # They're were also issues with pip not picking up some
+                # universal2 wheels, tag twice
                 additional_platforms.append("macosx_11_0_universal2")
 
             platform_tag_args = [f"--platform-tag=+{'.'.join(additional_platforms)}"]
@@ -107,7 +109,7 @@ def main():
         )
         (file,) = tmpdir.glob("*.whl")
         # unpack
-        proc = subprocess.run(["wheel", "unpack", file.name], cwd=tmpdir, check=True)
+        subprocess.run(["wheel", "unpack", file.name], cwd=tmpdir, check=True)
         for unpackdir in tmpdir.iterdir():
             if unpackdir.is_dir():
                 break
--- a/scripts/repair_wheel.py
+++ b/scripts/repair_wheel.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+
+import argparse
+import re
+import shutil
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+
+def main():
+    if sys.platform.startswith("linux"):
+        os_ = "linux"
+    elif sys.platform.startswith("darwin"):
+        os_ = "macos"
+    elif sys.platform.startswith("win32"):
+        os_ = "windows"
+    else:
+        raise NotImplementedError(
+            "sys.platform '{}' is not supported yet.".format(sys.platform)
+        )
+
+    p = argparse.ArgumentParser(
+        description="Convert wheel to be independent of python implementation and ABI"
+    )
+    p.set_defaults(prog=Path(sys.argv[0]).name)
+    p.add_argument("WHEEL_FILE", help="Path to wheel file.")
+    p.add_argument(
+        "-w",
+        "--wheel-dir",
+        dest="WHEEL_DIR",
+        help=('Directory to store delocated wheels (default: "wheelhouse/")'),
+        default="wheelhouse/",
+    )
+
+    args = p.parse_args()
+
+    file = Path(args.WHEEL_FILE).resolve(strict=True)
+    wheelhouse = Path(args.WHEEL_DIR).resolve()
+    wheelhouse.mkdir(parents=True, exist_ok=True)
+
+    with tempfile.TemporaryDirectory() as tmpdir_:
+        tmpdir = Path(tmpdir_)
+        # use the platform specific repair tool first
+        if os_ == "linux":
+            subprocess.run(
+                ["auditwheel", "repair", "-w", str(tmpdir), str(file)], check=True
+            )
+        elif os_ == "macos":
+            subprocess.run(
+                [
+                    "delocate-wheel",
+                    "--require-archs",
+                    "arm64,x86_64",
+                    "-w",
+                    str(tmpdir),
+                    str(file),
+                ],
+                check=True,
+            )
+        elif os_ == "windows":
+            # no specific tool, just copy
+            shutil.copyfile(file, tmpdir / file.name)
+        (file,) = tmpdir.glob("*.whl")
+
+        # we need to handle macOS universal2 & arm64 here for now, let's use platform_tag_args for this.
+        platform_tag_args = []
+        if os_ == "macos":
+            additional_platforms = []
+
+            # first, get the target macOS deployment target from the wheel
+            match = re.match(r"^.*-macosx_(\d+)_(\d+)_x86_64\.whl$", file.name)
+            assert match is not None
+            target = tuple(map(int, match.groups()))
+
+            # let's add universal2 platform for this wheel.
+            additional_platforms = ["macosx_{}_{}_universal2".format(*target)]
+
+            # given pip support for universal2 was added after arm64 introduction
+            # let's also add arm64 platform.
+            arm64_target = target
+            if arm64_target < (11, 0):
+                arm64_target = (11, 0)
+            additional_platforms.append("macosx_{}_{}_arm64".format(*arm64_target))
+
+            if target < (11, 0):
+                # They're were also issues with pip not picking up some universal2 wheels, tag twice
+                additional_platforms.append("macosx_11_0_universal2")
+
+            platform_tag_args = [f"--platform-tag=+{'.'.join(additional_platforms)}"]
+
+        # make this a py3 wheel
+        subprocess.run(
+            [
+                "wheel",
+                "tags",
+                "--python-tag",
+                "py3",
+                "--abi-tag",
+                "none",
+                *platform_tag_args,
+                "--remove",
+                str(file),
+            ],
+            check=True,
+        )
+        (file,) = tmpdir.glob("*.whl")
+        # unpack
+        proc = subprocess.run(["wheel", "unpack", file.name], cwd=tmpdir, check=True)
+        for unpackdir in tmpdir.iterdir():
+            if unpackdir.is_dir():
+                break
+        else:
+            raise RuntimeError("subdirectory not found")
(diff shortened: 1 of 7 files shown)


adamjstewart:
After this is merged we should consider cutting a 1.0.2 release so that we'll have wheels that can be installed on Python 3.12.

mwtoews:
The failures with `manylinux2014_i686.manylinux_2_17_i686` are puzzling, since there were no issues with the most-recent [build](https://github.com/Toblerity/rtree/actions/runs/5852463171). And for some reason the platform tags (separated by `.`) are flipped from `manylinux_2_17_i686.manylinux2014_i686`. Does the order of these matter? I'm able to reproduce these locally too, so they can be investigated further.

The macos wheels are also a bit odd. The previous build had two platform tags `macosx_10_9_x86_64` and `macosx_11_0_arm64` while the new has only one `macosx_12_0_arm64.macosx_12_0_universal2.macosx_12_0_x86_64`. Does this one wheel work for older platforms? I'm unable to do much testing with these.

mwtoews:
Oh wait, the previous builds worked for `manylinux*i686` since these tests were skipped with 67637889ecb44d6e57952eecdc3de4e023793169 . The skip could be re-instated, but I don't understand the root cause of the failures yet, they could be a 32-bit/64-bit thing.

Things are running oddly smooth for Windows, however! Additional platforms could be considered.

mwtoews:
I've enabled cp312 testing to see what happens. Currently testing is only possible on Linux, but all of the platforms show a consistent error:
```
  platform linux -- Python 3.12.0rc1, pytest-7.4.0, pluggy-1.2.0
  cachedir: .tox/py312/.pytest_cache
  rootdir: /project
  collected 38 items
  tests/test_index.py ........x............F...............                [ 97%]
  tests/test_tpr.py .                                                      [100%]
  =================================== FAILURES ===================================
  _____________________ IndexSerialization.test_interleaving _____________________
  self = <tests.test_index.IndexSerialization testMethod=test_interleaving>
      def test_interleaving(self) -> None:
          """Streaming against a persisted index without interleaving"""
      
          def data_gen(
              interleaved: bool = True,
          ) -> Iterator[Tuple[int, Tuple[float, float, float, float], int]]:
              for i, (minx, miny, maxx, maxy) in enumerate(self.boxes15):
                  if interleaved:
                      yield (i, (minx, miny, maxx, maxy), 42)
                  else:
                      yield (i, (minx, maxx, miny, maxy), 42)
      
          p = index.Property()
          tname = tempfile.mktemp()
          idx = index.Index(
              tname, data_gen(interleaved=False), properties=p, interleaved=False
          )
          hits1 = sorted(list(idx.intersection((0, 60, 0, 60))))
          self.assertTrue(len(hits1), 10)
          self.assertEqual(hits1, [0, 4, 16, 27, 35, 40, 47, 50, 76, 80])
      
  >       leaves = idx.leaves()
  /project/tests/test_index.py:475: 
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  self = rtree.index.Index(bounds=[-186.673789279, 184.761387556, -96.7177218184, 96.6043699778], size=100)
      def leaves(self):
          leaf_node_count = ctypes.c_uint32()
          p_leafsizes = ctypes.pointer(ctypes.c_uint32())
          p_leafids = ctypes.pointer(ctypes.c_int64())
          pp_childids = ctypes.pointer(ctypes.pointer(ctypes.c_int64()))
      
          pp_mins = ctypes.pointer(ctypes.pointer(ctypes.c_double()))
          pp_maxs = ctypes.pointer(ctypes.pointer(ctypes.c_double()))
          dimension = ctypes.c_uint32(0)
      
          core.rt.Index_GetLeaves(
              self.handle,
              ctypes.byref(leaf_node_count),
              ctypes.byref(p_leafsizes),
              ctypes.byref(p_leafids),
              ctypes.byref(pp_childids),
              ctypes.byref(pp_mins),
              ctypes.byref(pp_maxs),
              ctypes.byref(dimension),
          )
      
          output = []
      
          count = leaf_node_count.value
          sizes = ctypes.cast(p_leafsizes, ctypes.POINTER(ctypes.c_uint32 * count))
          ids = ctypes.cast(p_leafids, ctypes.POINTER(ctypes.c_int64 * count))
          child = ctypes.cast(
              pp_childids, ctypes.POINTER(ctypes.POINTER(ctypes.c_int64) * count)
          )
          mins = ctypes.cast(
              pp_mins, ctypes.POINTER(ctypes.POINTER(ctypes.c_double) * count)
          )
          maxs = ctypes.cast(
              pp_maxs, ctypes.POINTER(ctypes.POINTER(ctypes.c_double) * count)
          )
          for i in range(count):
              p_child_ids = child.contents[i]
      
              id = ids.contents[i]
              size = sizes.contents[i]
              child_ids_array = ctypes.cast(
                  p_child_ids, ctypes.POINTER(ctypes.c_int64 * size)
              )
      
              child_ids = []
              for j in range(size):
  >               child_ids.append(child_ids_array.contents[j])
  E               ValueError: NULL pointer access
  /project/rtree/index.py:1311: ValueError
  =============================== warnings summary ===============================
  tests/test_index.py::IndexSerialization::test_interleaving
    /project/rtree/index.py:283: DeprecationWarning: index.get_size() is deprecated, use len(index) instead
      warnings.warn(
  -- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
  =========================== short test summary info ============================
  FAILED tests/test_index.py::IndexSerialization::test_interleaving - ValueErro...
  ============== 1 failed, 36 passed, 1 xfailed, 1 warning in 0.62s ==============
```

mwtoews:
Finally ready for review. A few notes:

- Firstly, I'm newish to cibuildwheel, so review by anyone with more experience would be appreciated. Would appreciate @EwoutH to take a look too.
- There are many fewer built distribution files, e.g. linux went from 24 files 28.7 MB to 4 files 4.24 MB. This is a good thing.
- macOS now has a platform tag: `macosx_10_9_universal2.macosx_10_9_x86_64.macosx_11_0_arm64.macosx_11_0_universal2` which is better? Anyone with M1/M2 chip want to test this one?
- There is still an issue with cp312, these errors are shown in the workflow, but the tox config `ignore_outcome = True` allows the wheel builds to succeed
- This is still an issue with the i686 tests, but these don't seem to be major. The tests could be re-written to be more robust for 32-bit and 64-bits

adamjstewart:
> Anyone with M1/M2 chip want to test this one?

Me! Where can I download the wheel?

adamjstewart:
Would it be possible to have separate wheels for x86_64 and arm64 on macOS? We might see better performance if binaries aren't running through Rosetta.

mwtoews:
Go to the "Summary" of any build (e.g. [here](https://github.com/Toblerity/rtree/actions/runs/5920085484)), scroll to the bottom and find the "Artifacts", e.g. [macos-latest-whl](https://github.com/Toblerity/rtree/suites/15287878532/artifacts/873007461).

As for separate macos builds, sure I can take another look at the repair wheel script. As mentioned previously, I don't really understand much about the macos stuff, so advice would be appreciated.

adamjstewart:
I was able to import rtree. Anything else I should test?

I don't know much about cibuildwheel but I know some stuff about macOS. I'm going to guess that the macOS platform tag args are an artifact of when macOS arm64 first came out and no one had runners to properly build it so they instead built a single universal wheel. Are there any other repos that do something similar to this (single wheel for all Python versions) without universal wheels? I feel like there should be an easier way to do this than having a repair script... Maybe reach out to the cibuildwheel devs? 

@pradyunsg @rgommers may also have ideas on how to create platform-specific Python-version-independent wheels with cibuildwheel.

scripts/repair_wheel.py
```diff
@@ -0,0 +1,102 @@
...
+        if os_ == "linux":
+            subprocess.run(
+                ["auditwheel", "repair", "-w", str(tmpdir), str(file)], check=True
+            )
+        elif os_ == "macos":
+            subprocess.run(
+                [
+                    "delocate-wheel",
+                    "--require-archs",
+                    "arm64,x86_64",
```
adamjstewart, line 54: Not sure if we want this, this might be responsible for making a universal binary

mwtoews: I'll uncomment it for another build.

mwtoews:
Note the [previous macos artifacts](https://github.com/Toblerity/rtree/suites/15392516179/artifacts/878966641) have `macosx_10_9_x86_64` and `macosx_11_0_arm64` platform tags, but they are actually both x86_64 binaries. Testing again with CMAKE_OSX_ARCHITECTURES from AUDITWHEEL_ARCH variable...

mwtoews:
@adamjstewart mind testing the latest artifacts again? There are now two macOS wheels, each with only one architecture. Importing rtree is the first smoke test. But I suppose a performance benchmark would be more useful. There is no easy benchmark, unfortunately. The closest is [benchmarks.py.off](https://github.com/Toblerity/rtree/blob/32d3ae2b0cd846dc4fd25ca4970427ddd1f7dee9/tests/off/benchmarks.py.off) by @hobu

mwtoews:
Thanks all for testing and feedback! Last two things to mention.

1. The magic that rewrites the wheels is [`wheel tags`](https://wheel.readthedocs.io/en/stable/reference/wheel_tags.html), using `--python-tag=py3 --abi-tag=none` options. Additionally the `unpack` and `pack` steps are not really needed for Windows and macOS, but was there as a workaround for a (now closed) issue https://github.com/pypa/wheel/issues/558
2. The new wheels with "none" ABI tag work with PyPy, which supports ctypes. I could have added `pp{37,38,39}` to tox.ini, perhaps later.

PR #281, opened 2023-09-12
don't require file to be writeable if we're not going to write to it

### LINKED ISSUE #247: Index is checking write access despite assuming read-only access
# Context

When passing a filename to the Index class, it assumes that the file can not be overwritten unless explicitly set by the user. 
https://github.com/Toblerity/rtree/blob/81cc81a1f11081992bf2715bd82d08eca3e48e4e/rtree/index.py#L240-L244

However, passing a file which is in a read-only dir is not allowed, since it is checked if the provided file is within a directory with write access. 

https://github.com/Toblerity/rtree/blob/81cc81a1f11081992bf2715bd82d08eca3e48e4e/rtree/index.py#L255-L256

# Expected behavior

When the provided file is within a read-only directory, but overwrite `self.properties.overwrite = False` the no OSError should be raised.

jason-curtis:
closes #247

### DIFF
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -230,13 +230,13 @@ class Index:
             # check we can read the file
             f = str(basename) + "." + self.properties.idx_extension
             p = os.path.abspath(f)
-            path_exists = os.path.exists(p)
+            d = os.path.dirname(p)
+            file_exists = os.path.exists(f)
 
-            # assume if the file exists, we're not going to overwrite it
+            # assume we're not going to overwrite an existing file
             # unless the user explicitly set the property to do so
-            if path_exists:
-                self.properties.overwrite = bool(kwargs.get("overwrite", False))
-
+            self.properties.overwrite = bool(kwargs.get("overwrite", False))
+            if file_exists:
                 # assume we're fetching the first index_id.  If the user
                 # set it, we'll fetch that one.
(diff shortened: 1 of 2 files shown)


hobu:
Please include some tests so this behavior doesn't regress again.

jason-curtis:
Unfortunately it turns out the same check is happening on the C library level as well: https://libspatialindex.org/en/latest/doxygen/DiskStorageManager_8cc_source.html#l00173

I'm hoping it's an overzealous check and write access isn't actually required but I can't be sure. In any case I'm not equipped to do a quick C side project right now so I'm going to work around the issue instead. I've checked in my latest implementation including (failing) unit tests; feel free to close this PR or take it over if you prefer.

jason-curtis:
In fact, since the C code is already doing its own checks and has reasonable error messages, maybe we should just remove these checks entirely from the Python implementation.

mwtoews:
[repeating from linked issue] You could raise an issue and/or PR for libspatialindex; the source code on GitHub is [here](https://github.com/libspatialindex/libspatialindex/blob/3be67583852c741d0357f7ef5148c5fe8b484076/src/storagemanager/DiskStorageManager.cc#L173). But yeah, I agree that a fix from the library level would be better.

PR #286, opened 2023-10-03
Support and test Python 3.12

mwtoews:
Python 3.12 is out, so add "support" and test the version. Upgrade to cibuildwheel from today is for the Python 3.12.0 final release.

I'll suggest a micro release soon, so Python 3.12 folks can use the new built wheels.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
       fail-fast: true
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
         sidx-version: ['1.8.5','1.9.3']
 
     steps:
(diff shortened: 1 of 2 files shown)


adamjstewart (commented):
Should we also bump all style tests and readthedocs config from 3.11 to 3.12?

mwtoews:
I didn't because I'm not sure if all the upstream dependencies are ready for Python 3.12 yet. E.g. [here](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-tools-python).

adamjstewart (approved):

EwoutH:
Thanks for this effort!

Would it be possible to tag a new release to get Python 3.12 wheels up on PyPI? Currently only Python 3.11 wheels are available as part of the 1.0.1 release.

The current version of cibuildwheel already builds wheels for Python 3.12, so this should go (almost) fully automatic.

mwtoews:
actually cibuildwheel was reconfigured to build one set of wheels for Python >=3.8, since this codebase is effectively pure Python with platform dependant shared libs via ctypes.

@hobu are you able to create a release?

hobu:
@mwtoews gave you Admin to be able to snap off the release whenever you wish. I haven't been tracking all of this stuff very closely.

mwtoews:
In case you missed it @EwoutH see new release https://pypi.org/project/Rtree/1.1.0/#files

EwoutH:
Awesome, thanks!

PR #287, opened 2023-10-03
Upgrade typing for tests (via pyupgrade)

mwtoews:
This was mostly automated with pyupgrade. Review appreciated as I'm a typing novice.

Note that these files have `from __future__ import annotations` at the top; xref #273

### DIFF
--- a/tests/test_index.py
+++ b/tests/test_index.py
@@ -4,7 +4,7 @@ import ctypes
 import pickle
 import tempfile
 import unittest
-from typing import Dict, Iterator, Tuple
+from typing import Iterator
 
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


adamjstewart:
FWIW, I stopped using `from __future__ import annotations` in my code because it hasn't yet been adopted into the official standard and it doesn't seem like they're close to reaching a decision: https://docs.python.org/3/library/__future__.html

Downside is that the updated type hints require Python 3.10 to run mypy. Upside is they're much prettier. We could also consider dropping support for older Python, but I don't see any reason to rush that.

Basically, I'm not opposed to this change, but I also don't think it's necessary. As long as we're consistent across all files I'm happy.

mwtoews:
Thanks for your incite @adamjstewart ! I'll merge for now, since the scope is just for tests. It was also part of some pre-commit scoping that I'm considering to add.

PR #288, opened 2023-10-15
Add pre-commit, replace GHA workflows, remove flake8 config

mwtoews:
This also uses [ruff](https://github.com/astral-sh/ruff) in-place of flake8, so clean up a now unused `.flake8` file.

Several of the GHA jobs are also redundant, and cleaned-up.

The maintenance-only [pre-commit/action@v3.0.0](https://github.com/pre-commit/action) is chosen (for now), but should ideally be using [pre-commit.ci](https://pre-commit.ci/) (which I don't have permissions to configure).


.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: check-yaml
+    -   id: end-of-file-fixer
+    -   id: trailing-whitespace
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.13.0
+    hooks:
+    -   id: pyupgrade
```
adamjstewart, line 11: Doesn't ruff also replace pyupgrade and isort? https://github.com/astral-sh/ruff#rules

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: check-yaml
+    -   id: end-of-file-fixer
+    -   id: trailing-whitespace
```
adamjstewart, line 7: Didn't know about these, neat!

adamjstewart (commented):
I'm not the biggest fan of pre-commit/pyupgrade but they are pretty popular and reasonable

.pre-commit-config.yaml
mwtoews: Good point. I'll revise.

mwtoews:
> I'm not the biggest fan of pre-commit/pyupgrade but they are pretty popular and reasonable

Thanks for weighing in @adamjstewart as I thought it was an annoying tool for a while tool. I now see it's value, and indeed has become a popular tool across many projects.

adamjstewart (approved):
I prob won't use it directly, but I'm fine with it being used in CI.

PR #292, opened 2023-11-08
Include spatialindex headers

JDBetteridge:
It would be great if the spatialindex headers could be included with the packaged binary wheel of Rtree. We have two  projects that currently depend on spatialindex:
- [Firedrake](https://github.com/firedrakeproject/firedrake), which links to `libspatialindex` when performing code generation and we hope to in the future transition to using the Python interface that Rtree provides.
- [Supermesh](https://github.com/firedrakeproject/libsupermesh), which also links to `libspatialindex`, but is minimally maintained for users who depend on the functionality that this library provides.

We would like to move away from building our own copy of spatialindex (twice!) to linking against the shared library packaged with Rtree. To be able to build the projects all we require are the spatialindex headers and the introspection functions that this PR provides.

For reference you can see how we would accomplish this in these PRs:
- https://github.com/firedrakeproject/firedrake/pull/3138
- https://github.com/firedrakeproject/libsupermesh/pull/4

Please let me know your thoughts and suggestions :slightly_smiling_face:

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,4 +3,3 @@ include MANIFEST.in
 recursive-include benchmarks *
 recursive-include tests *
 recursive-include docs *
-recursive-include rtree/include *.h
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,3 +3,4 @@ include MANIFEST.in
 recursive-include benchmarks *
 recursive-include tests *
 recursive-include docs *
+recursive-include rtree/include *.h
--- a/ci/install_libspatialindex.bash
+++ b/ci/install_libspatialindex.bash
@@ -7,11 +7,21 @@ SHA256=63a03bfb26aa65cf0159f925f6c3491b6ef79bc0e3db5a631d96772d6541187e
 
 # where to copy resulting files
 # this has to be run before `cd`-ing anywhere
-gentarget() {
+libtarget() {
   OURPWD=$PWD
   cd "$(dirname "$0")"
   mkdir -p ../rtree/lib
   cd ../rtree/lib
+  arr=$(pwd)
+  cd "$OURPWD"
+  echo $arr
+}
+
+headertarget() {
+  OURPWD=$PWD
+  cd "$(dirname "$0")"
+  mkdir -p ../rtree/include
+  cd ../rtree/include
   arr=$(pwd)
   cd "$OURPWD"
   echo $arr
@@ -26,7 +36,8 @@ scriptloc() {
 }
 # note that we're doing this convoluted thing to get
 # an absolute path so mac doesn't yell at us
-TARGET=`gentarget`
+LIBTARGET=`libtarget`
+HEADERTARGET=`headertarget`
 SL=`scriptloc`
 
 rm $VERSION.zip || true
@@ -60,10 +71,13 @@ if [ "$(uname)" == "Darwin" ]; then
     # change the rpath in the dylib to point to the same directory
     install_name_tool -change @rpath/libspatialindex.6.dylib @loader_path/libspatialindex.dylib bin/libspatialindex_c.dylib
     # copy the dylib files to the target director
-    cp bin/libspatialindex.dylib $TARGET
-    cp bin/libspatialindex_c.dylib $TARGET
+    cp bin/libspatialindex.dylib $LIBTARGET
+    cp bin/libspatialindex_c.dylib $LIBTARGET
+    cp -r ../include/* $HEADERTARGET
 else
-    cp -d bin/* $TARGET
+    cp -L bin/* $LIBTARGET
+    cp -r ../include/* $HEADERTARGET
 fi
 
-ls $TARGET
+ls $LIBTARGET
+ls -R $HEADERTARGET
--- a/ci/install_libspatialindex.bat
+++ b/ci/install_libspatialindex.bat
@@ -21,7 +21,7 @@ ninja
 
 mkdir %~dp0\..\rtree\lib
 copy bin\*.dll %~dp0\..\rtree\lib
-copy ..\/include/* %~dp0\..\rtree\include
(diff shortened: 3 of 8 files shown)


JDBetteridge:
All the CI checks appear to be passing now! Please let me know if you have any requested changes or would like me to tidy the git history, etc

JDBetteridge:
I'm not saying it 100% justifies the practise, but `numpy` do this, for one example. (`import numpy; numpy.get_include()`)

mwtoews:
Thanks for the PR! As for the analog with NumPyh, I see headers in `numpy/core/include/numpy` for both their sdist (source distribution) but also their bdist (built distribution) on PyPI.

As for this PR, I see two behaviors:

1. If libspatialindex is built and installed (e.g. via `ci/install_libspatialindex.*` scripts), then the headers are included for both sdist and bdist (like NumPy on PyPI)
2. If libspatialindex is not installed, then headers are absent from sdist and bdist

Is this expected? (I think so, but just checking).

JDBetteridge:
Not wanting to be impatient, just want to check the status of this PR :slightly_smiling_face: 

I think we are in very different time zones, so I appreciate communication might take more time!

JDBetteridge:
Thanks for the feedback, I will address this as soon as I get a chance. In the sdist case I imagine that the best thing for the `get_include()` introspection function to return is an empty string.

JDBetteridge:
I realise that fixing this requires minimal changes. Because the `sdist` is created in a separate CI step ("Package and push release") in an isolated environment, you already get the desired behaviour. Looking in the previous logs, you can see when building the `sdist` a warning is raised `warning: no files found matching '*.h' under directory 'rtree/include'`. I believe that if you were to additionally upload the `sdist` artifact (which the below patch would do, but I do not include it as a commit in case it is undesirable behaviour!) you would see that the `include` directory is absent.

```patch
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
index e43b341..274d8e8 100644
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -138,6 +138,10 @@ jobs:
             python3 -m pip install setuptools numpy pytest wheel
             export PATH=$PATH:/home/runner/.local/bin
             python3 setup.py sdist
+      - uses: actions/upload-artifact@v3
+        with:
+          name: sdist
+          path: dist/*.tar.gz
 
       - uses: actions/download-artifact@v3
         with:
```

Regardless, I have fixed the `get_include()` function to return an empty string if running as part of an `sdist` installation. This is consistent with the behaviour of `get_libraries()`.

JDBetteridge:
Sorry if I didn't make this obvious, but I think I'm done here! Polite bump :slightly_smiling_face:

mwtoews:
Hi yes, I'm happy to go forward with some version of this PR, but will still need some time to do a more thorough review, so appreciate the time. I hope to find a bit more time over the week.

mwtoews:
Note that I've not forgotten this PR, and have a few local edits that I've made that I'll push here when I'm ready.

JDBetteridge:
Awesome! Thanks for the update :slightly_smiling_face:

mwtoews:
Ok, back on this for more questions.

What is the purpose of `get_libraries()`, and who would use it? It seems a bit out of place. If anyone really needs it, I suppose they could just do this:
```python
import rtree
rtree.core.rt._name  # /path/to/lib/libspatialindex_c.so.6
```
As for `get_include()`, this mirrors nicely with [`numpy.get_include()`](https://numpy.org/devdocs/reference/generated/numpy.get_include.html).  But could I move it to `rtree.finder.get_include()`? I have a few adjustments to make it work for Python 3.8 and installs that get `libspatialindex` via conda-forge or Debian-like users with `libspatialindex-dev` installed. I also would like to add a bit of testing.

mwtoews:
Back to @JDBetteridge for review. A summary of relevant changes:

- Moved `get_include()` to `rtree.finder` because it is a more appropriate module
- Adapted it to search for headers in priority from (1) binary wheel (using `importlib.metadata`), (2) `sys.prefix` for conda, (3) `/usr` for system package install (e.g. [`libspatialindex-dev`](https://packages.ubuntu.com/focal/libspatialindex-dev)).
- Removed `rtree.core.get_libraries()` and `rtree.core.get_library_name()` since this a bit out of scope

Some other notes is that a binary wheel with lib and include is created via `python -m build --wheel`, but not with `python -m build`, because the later command builds the wheel from the sdist. I've reworked this PR to exclude lib and include from the sdist.

The built wheel artefacts can be found [here](https://github.com/Toblerity/rtree/actions/runs/7550871663), which can be downloaded unzipped and installed via pip.

Any feedback welcome!

JDBetteridge:
Thanks @mwtoews!

To answer your first question, I added `get_libraries()` mainly as a convenience function for our use case, which is to compile small C kernels using symbols from the spatialindex shared library. However, I will concede that this is a fairly niche use case and I'm happy to use your suggestion of obtaining the path and shared library from the private attribute `rt._name`.

- I agree that `rtree.finder` is a better place for the `get_include()` function.
- I am not a *conda user, but the header priority appears correct. Just to double check: Your intention with this priority is to have users who install the binary wheel to use the bundled `libspatialindex_c.so` over any system spatialindex, unless they set the `SPATIALINDEX_C_LIBRARY` environment variable? I don't disagree with the choice, I just want to check I have understood correctly.
- My key aim for this PR was to get the headers bundled with the wheel, the library introspection functions can be omitted.

I'm very happy for this PR to be merged in its current state. I have run our CI on this branch today and everything is working as expected.

mwtoews:
The environ `SPATIALINDEX_C_LIBRARY` can be used to load a specified shared library at runtime, so yes you are understanding this correctly. It's mostly a developer feature, or where multiple instances of the library are available.

Thanks for testing and feedback!

Issue #105, opened 2018-04-20
Multithreading issue


yatagan:
My code works well until I start to use several threads:

```
<class 'ctypes.ArgumentError'>: argument 5: <class 'TypeError'>: expected LP_LP_c_void_p instance instead of pointer to LP_c_void_p
```

It happens on line 481: https://github.com/Toblerity/rtree/blob/483bd9b21d897a9cc63642517c24d637f008b1bd/rtree/index.py#L481

Rtree==0.8.3

yatagan:
#53

hobu:
Rtree  is not threadsafe. Patches accepted to improve the situation, but know that it would be a lot of work.

Issue #53, opened 2015-07-01
Multithreading


christophlingg:
Thanks for providing RTree for Python!

I was wondering if rtree can be used in a multithreaded context?

shongololo:
+1

marfago:
Actually seems it is not possible. At least with recent versions.

https://github.com/libspatialindex/libspatialindex/issues/71

hobu:
libspatialindex is not threadsafe and Rtree does not do thread safety. It is something that could be done, but it would take a lot of meticulous work to achieve.

christophlingg:
Ok I can comprehend that concurrent writing and reading may cause problems, but are read-only accesses safe?

marfago:
In my experience it seems it is not.
In a project I use https://github.com/collective/collective.geo.index, wich is based on RTree. For each search it instantiates a new instance of RTree and now and then it crashes with segmentation fault. I solved this issue moving the rtree to a global variable (just one instance). Not sure it can help.

christophlingg:
That's strange because I am using accessing one rtree from multiple threads, and so far i did not encounter any problems.

snorfalorpagus:
@christophlingg Race conditions can be like that. It may be that certain issues are only triggered under high(er) loads.

hobu:
> Ok I can comprehend that concurrent writing and reading may cause problems, but are read-only accesses safe?

No. libspatialindex, especially as used through the C API like Rtree does, is not thread safe. If it is working for you, it is most likely due to luck.

christophlingg:
Now I know I need to fix something ;-) Thanks to your feedback this I can do this before it crashes!

hobu:
The most pressing issue is probably the one described in libspatialindex/libspatialindex#71 especially in regards to the C API.

If you put some aggressive locking around things in python land, you are likely to be reasonably protected. The issues most likely manifest themselves in regard to error messages, and the error stack of the C API of libspatialindex not being protected in any way.

marfago:
I'm working with a single thread instance of the index, but it daily crashes!
Can anyone suggest what exactly should be "protected" in python?

hobu:
> Can anyone suggest what exactly should be "protected" in python?

Protect any access you do to libspatialindex with a mutex and lock accordingly.

marfago:
@hobu :Adding mutex or lock did not work for me. Moreover, also using different processes working on the same index file ends up in segmentation fault.

hobu:
[...]

PR #307, opened 2024-02-13
Docs updates - multithreading/process recommendations + strikethrough of disk serialization

codeananda:
Given the multithreading/processing issues with rtree and libspatialindex, I thought it made sense to add this info to the README (and also the online docs). 

Related discussions that inspired my updates:

* https://github.com/libspatialindex/libspatialindex/issues/71
* #53 
* #105

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ user.  These features include:
 * Clustered indexes (store Python pickles directly with index entries)
 * Bulk loading
 * Deletion
-* Disk serialization
+* ~~Disk serialization~~ [currently broken as of Jan 2024](https://github.com/Toblerity/rtree/pull/197)
 * Custom storage implementation (to implement spatial indexing in ZODB, for example)
 
 
--- a/README.md
+++ b/README.md
@@ -17,6 +17,19 @@ user.  These features include:
 * ~~Disk serialization~~ [currently broken as of Jan 2024](https://github.com/Toblerity/rtree/pull/197)
 * Custom storage implementation (to implement spatial indexing in ZODB, for example)
 
+These features do not include:
+
+* Multithread safety (for reading or writing)
+* Multiprocess safety (for reading or writing)
+
+For either of these, we recommend using:
+
+* A PostGIS database, or
+* GeoPandas + spatial joining, or
+* Building an rtree from scratch in each thread/process (using [generator syntax](https://rtree.readthedocs.io/en/latest/performance.html#use-stream-loading))
+
+Note that since rtree is written in C, it can be orders of magnitude faster than a 
+database even if running sequentially.
 
 Wheels are available for most major platforms, and `rtree` with bundled `libspatialindex` can be installed via pip:
 


codeananda:
Is your issue that I've linked to the wrong PR/issue? Or more that we're updating the docs instead of fixing the problem? I can help with the former but not sure I know how to fix the latter.

hobu:
> Or more that we're updating the docs instead of fixing the problem?

Correct. Is there a ticket that clearly demonstrates the issue? I don't think we can categorically say disk serialization doesn't work. Maybe it doesn't work for you and your situation, but we would be flooded with tickets if it was broken more generally.

adamjstewart:
I know pickling doesn't work (which means multiprocessing doesn't work on macOS/Windows): #87

hobu:
> I know pickling doesn't work

pickling isn't the same as "supports disk serialization". It's always been possible to use Rtree to store whatever data you wanted to with an index entry. That it's not conveniently pickled python objects is orthogonal.

codeananda:
> pickling isn't the same as "supports disk serialization"

I didn't know this. 

The docs give the impression that pickling is synonymous with disk serialisation. Indeed, so does the code. 

1. The [Serializing your index to a file](https://rtree.readthedocs.io/en/latest/tutorial.html#serializing-your-index-to-a-file) section doesn't actually tell you how to serialize to a file, but rather read an already created file from disk. 
2. The only section in [performance](https://rtree.readthedocs.io/en/latest/performance.html#override-dumps-to-use-the-highest-pickle-protocol) relating to disk serialization recommends using `cpickle`.
3. The `Index` class has `dumps` and `loads` methods - both one-liners that use `pickle.loads/dumps`
4. `Index._serialize` calls `self.dumps` on the first line.

I genuinely thought the only way to serialize rtree objects to disk was pickle.

Or am I not understanding the meaning of "disk serialization"?

At the very least, we need to add a snippet under [Serializing your index to a file](https://rtree.readthedocs.io/en/latest/tutorial.html#serializing-your-index-to-a-file) that shows how you write to disk.

codeananda:
@hobu @mwtoews any update?

FreddieWitherden:
> At the very least, we need to add a snippet under [Serializing your index to a file](https://rtree.readthedocs.io/en/latest/tutorial.html#serializing-your-index-to-a-file) that shows how you write to disk.

I believe you can pass `None` for the data and still have an R-tree stored on disk.  This is *probably* what most people think of in terms of disk serialization.  Of course, here, you are on your own in terms of associating Python objects with each entry (just going by their integer id's).  This, I believe, is a better option and likely more efficient (and closer to what databases do: indices store pointers to the data rather than the data itself).  Keeps your index small and enables you to change your data more freely and have multiple pointers to the same data block.

hobu:
#423 removes pickle storage, so maybe this ticket is no longer relevant. reopen it with new information or demonstration of any significant issues.

adamjstewart:
This is a different thing. We removed object pickling, but the index itself should be pickleable and is still broken.

Issue #315, opened 2024-06-10
1.3.0 release?


hobu:
@mwtoews with the release of libspatialindex 2.0.0, the current release of Rtree has tests that are broken (a version check). We should issue a new Rtree release.

Did you have any items you wanted to make sure were addressed or have any time to work some of the issues before we snapped off the release?

adamjstewart:
If there are no new features/API changes and only bug fixes, shouldn't it be 1.2.1?

hobu:
> shouldn't it be 1.2.1?

No strong opinion but given our release cadence being so slow, I thought a major version number would also align with prompting an update to libspatialindex since they are so closely linked.

Our wheel generation stuff would also need to be updated to do libspatialindex 2.0.0

mwtoews:
Unfortunately, I have limited time to address new features / resolve [issues](https://github.com/Toblerity/rtree/issues) with Rtree. However, I can find some time this week to update the wheel generation for libspatialindex 2.0.0 and push out a new release.

As for the next logical version, I don't have any strong opinions. Since libspatialindex headers are included with binary wheels, probably a minor increment to 1.3.0 makes most sense.

PR #316, opened 2024-06-11
CI/RLS: upgrade to libspatialindex-2.0.0

mwtoews:
Towards #315.

This PR does the following:

- Upgrade libspatialindex from 1.9.3 to 2.0.0
- Remove the unneeded ci/CMakeLists.txt
- Use cmake install, removing the unneeded lib subdirs
- Remove the soversion number and symlinks for the library name (via [`CMAKE_PLATFORM_NO_VERSIONED_SONAME=ON`](https://cmake.org/cmake/help/latest/variable/CMAKE_PLATFORM_NO_VERSIONED_SONAME.html))
- It turns out the cibuildwheel tests on different python versions (via tox) was misleading, as it was not testing the built-wheel but the local source tree and sidx library build. This is fixed with `pytest --import-mode=importlib` 
- Consequently, the musllinux wheels on PyPI are "broken", as verified with `docker run -it --rm alpine:3.18 /bin/sh --login` then `apk add py3-pip`, `pip install rtree`, `python -c "import rtree"` to observe error. This is resolved by setting rpath for libspatialindex described next.
- Use `CMAKE_INSTALL_RPATH` only for macOS. On Linux, `$ORIGIN` seems to confuse auditwheel and prevents creation of `Rtree.libs`. It want's LD_LIBRARY_PATH set while repairing the wheel. But the `$ORIGIN`  rpath is still needed for musllinux, so manually add this last step in repair_wheel.py
- Simplify copying directory trees in setup.py with [copy_tree](https://github.com/pypa/distutils/blob/4e6d97db1b4364cb184960f477c3e78435332646/distutils/cmd.py#L364)


hobu:
Waiting on you to merge this @mwtoews. Should we fire a release afterward?

mwtoews:
there one thing that I'm sorting out before merge, as I want to keep the linux lib dir in Rtree.libs, since I see downstream packages rely on this.

mwtoews:
There is yet another big list of changes on the build-side, updated in the PR description. This is ready to review by anyone, I'll probably merge it in another day after I've had a chance to carefully review the artefacts.

EwoutH:
This looks like an amazing clean-up!

One thing I noticed, is that in the build CI [vcpython27](https://community.chocolatey.org/packages/vcpython27) is installed on Windows:

https://github.com/Toblerity/rtree/blob/d42650ccd78a413ce016a8528a859ed8ab20c1ee/.github/workflows/deploy.yml#L43-L47

Is this still needed, or can it be updated?

mwtoews:
It looks like vcpython27 was added in 2020 with #174, at the time when MSBuild was used. Perhaps the existing ilammy/msvc-dev-cmd is sufficient, since cl.exe is needed for `cmake -G Ninja` (in ci/install_libspatialindex.bat).

PR #319, opened 2024-07-09
Update code style, replace isort and black with ruff, modern numpy rng

mwtoews:
Update the Python code style (again!) using ruff lint tools, which replace isort and black. Also select pyupgrade and numpy-specific checks. The Python code changes are mostly style, and should not have any functional changes.

The updated pre-commit hooks are also integrated with pre-commit.ci

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,25 +1,27 @@
+ci:
+  autoupdate_schedule: quarterly
 repos:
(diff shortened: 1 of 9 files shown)


adamjstewart:
Big fan of ruff, I'm using it in TorchGeo too! Our settings for reference: https://github.com/microsoft/torchgeo/blob/main/pyproject.toml#L262

Issue #223, opened 2022-02-21
Port bindings to pybind11?


musicinmybrain:
It was suggested by @hobu in https://github.com/Toblerity/rtree/issues/220#issuecomment-1046258628 that the [libspatialindex](https://github.com/libspatialindex/libspatialindex) bindings should ideally be ported from [ctypes](https://docs.python.org/3/library/ctypes.html) to [pybind11](https://pybind11.readthedocs.io/en/stable/index.html).

This could reduce the need for manual alignment of function prototypes and manual type hinting, and it could hopefully prevent future subtle bugs due to prototype mismatches (cf. https://github.com/Toblerity/rtree/issues/220).

It would, however, require new machinery to build and publish binary wheels for popular platforms.

I am filing this issue to provide a forum for discussion and investigation, and to track eventual progress, on this topic.

musicinmybrain:
Users seem to expect binary wheels, not just `sdist`-only releases, for PyPI packages with compiled code. I’ve seen a few options for building binary wheels for assorted platforms. I haven’t deployed any of them myself. It seems like [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) is a popular and relatively polished option. As is typical in Python packaging, there are almost as many workflows, processes, and combinations of tools as there are projects.

Integrating something like this into the release process will probably take some effort and decision-making by the project owners.

musicinmybrain:
Comparison with selected other tools for wrapping C APIs:

- [cffi](https://cffi.readthedocs.io/en/latest/index.html) is pleasantly simple and lightweight, but manual definition of function prototypes is still required, and mismatches are still possible. It is only suitable for wrapping C APIs; in this case, that would be fine. It’s mostly aimed at low-level bindings, akin to the current `ctypes` binding, and it’s usually possible to implement these with the interface definitions and little or no additional C code.
- [Cython](https://cython.org/) offers a powerful general-purpose almost-Python language that is transpiled to C or C++. It can be particularly useful for high-level bindings that need to implement significant functionality in compiled code for performance reasons, especially when a goal is to minimize the amount of C or C++ code needed. Unlike `cffi`, it can wrap C++. Manual definition of function prototypes is still required, and mismatches are still possible. Updates to Cython aren’t always entirely backwards-compatible, and its language isn’t formally specified as C or C++ are.
[...]

PR #321, opened 2024-07-10
Drop Python 3.8 support, require Python 3.9+

EwoutH:
With 1.3.0 freshly released this seems like a good moment to drop Python 3.8 support. It's almost [end-of-life](https://endoflife.date/python) and [SPEC 0](https://scientific-python.org/specs/spec-0000/) already recommended dropping Python 3.8 over a year ago.

In fact, they even recommend already dropping Python 3.9 and requiring 3.10+, which many packages do. But that might be a bit much for one PR.

So this PR bumps the required Python version to 3.9, removes 3.8 from the CI and test files and applies [pyupgrade](https://github.com/asottile/pyupgrade) with `--py39-plus`.

Of course this will only affect future releases, existing releases stay compatible with Python 3.8 and can continued to be used.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
       fail-fast: true
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
         # test oldesst and newest libspatialindex versions
         sidx-version: ['1.8.5', '2.0.0']
         exclude:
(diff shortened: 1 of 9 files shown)


adamjstewart:
I agree with dropping support for Python 3.8, although I think we should officially decide whether we're following the CPython support schedule, NEP 29 (superseded), SPEC 0, or a different support schedule.

Note that we're no longer using pyupgrade, we now use ruff instead.

mwtoews:
I'm inclined to follow the CPython / PEP release schedule shown [here](https://devguide.python.org/versions/), e.g. [PEP 569](https://peps.python.org/pep-0569/) for Python 3.8. This project is _mostly_ pure Python, so it isn't too much trouble to support 4 to 5 Python versions at the same time. But say if, for example, #223 were implemented to use pybind11 then SPEC 0 would make more sense to limit the number of binary wheel combinations.

mwtoews (approved):
Changes look good. And if we all agree to the CPython / PEP release schedule, then the timing of this change is good too.

PR #322, opened 2024-07-10
CI: Add test job for Python 3.13

EwoutH:
Python 3.13 is now in beta, thus is might be useful to start testing what works and what doesn't.

This PR adds a Python 3.13 jobs for the conda and regular Ubuntu pip jobs.

It also fixes a bug in the Ubuntu job, by making sure the Python version specified in the matrix is actually used when setting up Python.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
       fail-fast: true
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
         # test oldesst and newest libspatialindex versions
         sidx-version: ['1.8.5', '2.0.0']
         exclude:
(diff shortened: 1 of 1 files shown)


EwoutH:
Looks like it's stuck on https://github.com/samuelcolvin/watchfiles/issues/276. Waiting on https://github.com/samuelcolvin/watchfiles/pull/287 to get merged and a new watchfiles release to be published.

.github/workflows/test.yml
```diff
@@ -63,14 +63,15 @@ jobs:
...
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
 
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v5
       name: Install Python
       with:
-        python-version: '3.11'
+        python-version: ${{ matrix.python-version }}
```
adamjstewart, line 73: Can't believe we weren't actually using this before 🤦

adamjstewart: At least the conda tests didn't have this bug, so it's not like we were never testing all Python versions.

mwtoews: Nice catch! Amazing that this one has slipped by, taking a look at the former build.yml this has persisted a few years.

mwtoews (approved):
Thanks! Tests with 3.13b2 look good, no new errors or warnings. We'll add it to tox workflow after the final release in October.

EwoutH:
Thanks for merging!

> We'll add it to tox workflow after the final release in October.

I recommend doing this after release candidate 1, which is planned for Tuesday, 2024-07-30.

PR #325, opened 2024-07-11
CI: fail with warnings, and several minor tunings to test workflow

mwtoews:
Notes on the changes:

- Tests will fail from warnings (`-Werror`)
- Use verbose messages with pytest, since this is the primary output to review
- Change `fail-fast: false` to run all jobs of a matrix, even if there is one failure
- Remove the pre-commit workflow job, since pre-commit.ci was recently configured
- Use conda to install pytest for the conda jobs (not pip)
- Apt should't install python3-pip, and use `apt-get` for non-interactive use
- Remove 'paths' for 'push', so that the tests are re-run after pull-requests are merged to master

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -51,7 +51,7 @@ jobs:
 
     - uses: actions/upload-artifact@v4
       with:
-        name: cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}
+        name: cibw-wheels-${{ matrix.os }}
         path: ./wheelhouse/*.whl
 
   build_sdist:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,21 +4,12 @@ on:
   push:
     branches:
       - master
-    paths:
-      - '.github/workflows/test.yml'
   pull_request:
   workflow_dispatch:
   schedule:
     - cron: '0 6 * * 1'
 
 jobs:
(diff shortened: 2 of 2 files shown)


EwoutH (approved):
Looks good!

.github/workflows/deploy.yml
```diff
@@ -51,7 +51,7 @@ jobs:
 
     - uses: actions/upload-artifact@v4
       with:
-        name: cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}
```
EwoutH, line 54: I don't understand what's happening here fully, I would make sure to test this.

mwtoews: The `-${{ strategy.job-index }}` was an necessary number for the artefact name, e.g. "cibw-wheels-macos-latest-2.zip". So now it becomes just "cibw-wheels-macos-latest.zip". Artefacts can be viewed/downloaded in the GHA Summary, which I sometimes download to inspect or upload to PyPI (hopefully less-so for future releases).

EwoutH: Doesn't this give a problem with macOS artifact names? Since cibuildwheel runs both on x86 and Arm runners?

mwtoews: I just checked to see, and the files look good. Each artefact contains several wheel files created from the runner, so there's no conflict.

PR #326, opened 2024-07-11
CI: Always run on push to master branch

EwoutH:
Currently the CI doesn't run when a commit is pushed to the master branch, except if the configuration itself is changed. This is a bit strange, because you would always want to validate that the master branch is testing properly, especially with more complex squash / rebase merges.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - master
-    paths:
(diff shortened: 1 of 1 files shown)


mwtoews:
I saw this too, and already have #325 that does the same as this PR.

EwoutH:
Didn't saw that one, closing in favor of #325!

PR #327, opened 2024-07-11
Update .readthedocs.yaml: Use latest Ubuntu and Python versions

EwoutH:
Use latest Ubuntu and Python versions for Readthedocs. These are faster and throw more detailed errors if anything goes wrong.


mwtoews (approved):
Thanks! It looks like a seamless build.

PR #328, opened 2024-07-12
CI/BLD: move CIBW commands to pyproject.toml, move scripts

mwtoews:
This PR does the following:

- Move a few cibuildwheel commands from the CI workflow to pyproject.toml
- Remove unneeded vcpython27 package installation for Windows (xref [this comment](https://github.com/Toblerity/rtree/pull/316#issuecomment-2222557286))
- Move a few scripts from ci/ to /scripts
- The install_libspatialindex.bash script is really just a POSIX shell (i.e. no bash features), so rename and use `sh`


scripts/install_libspatialindex.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
adamjstewart, line 1: For fun, I tried running this script with dash on macOS. I confirmed that it's POSIX compliant. The only issue I encountered is that `sha256sum` does not exist on macOS, you have to use `shasum -a 256` instead. Might be able to check `uname` in an if-statement if you want to fix this. I can also submit a PR after this one is merged if you want.

mwtoews: Thanks for checking! It looks like the macos image has `sha256sum`, probably via brew's coreutils package (the internet tells me at least). But I won't mind if you want to complete this aspect. You can push commits on this PR, if you know how.

adamjstewart: > I confirmed that it's POSIX compliant.

I take this back. `==` is bash-only, we should be using `=` instead. I'll fix this in my commit.

adamjstewart: Btw, https://www.shellcheck.net/ is handy if you want to check for POSIX compliance and other sh gotchas.

mwtoews: Handy resource! The cli version does nice work too.

PR #340, opened 2024-12-07
Add support for array-based bulk insert

FreddieWitherden:
This is the counterpart to https://github.com/libspatialindex/libspatialindex/pull/267

### DIFF
--- a/rtree/core.py
+++ b/rtree/core.py
@@ -135,7 +135,7 @@ try:
         ctypes.c_uint64,
         ctypes.c_void_p,
         ctypes.c_void_p,
-        ctypes.c_void_p
+        ctypes.c_void_p,
     ]
     rt.Index_CreateWithArray.restype = ctypes.c_void_p
     rt.Index_CreateWithArray.errcheck = check_void  # type: ignore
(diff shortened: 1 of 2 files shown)


FreddieWitherden:
Test failures are expected as it depends on new functionality for libspatialindex.  For my case of inserting ~1 million points time goes from ~6s to ~1s.

hobu:
These tests should be resilient to different versions of libspatialindex. Rtree releases are not pinned to specific libspatialindex releases.

FreddieWitherden:
> These tests should be resilient to different versions of libspatialindex. Rtree releases are not pinned to specific libspatialindex releases.

Have updated the PR so we catch the exceptions.  Does this approach seem reasonable (assuming it checks out test wise?)

FreddieWitherden:
Just a short note about the implementation: the inputs for this are required to be NumPy arrays however I've been careful to not import NumPy so it is not a dependency of the library.  This should be reasonably flexible and anyone looking to seriously take advantage of the bulk capabilities really needs to be using NumPy (or a compatible alternative) anyway.

PR #341, opened 2024-12-14
Consistency checks

FreddieWitherden:
At the moment attempting to do something like:
```
props = Property(dimension=3, index_capacity=25, leaf_capacity=25)
idx = Index([], properties=props)
```
will throw an exception because the near minimum overlap factor is greater than the leaf/index capacities.  Interestingly (and for reasons I can not understand) it does not throw an exception if `[]` is replaced by actual entries.  It is only an issue for an initially empty index.

Either way, this PR adds some consistency checks so everything works as expected.

### DIFF
--- a/rtree/core.py
+++ b/rtree/core.py
@@ -238,6 +238,44 @@ rt.Index_NearestNeighbors_id.argtypes = [
 rt.Index_NearestNeighbors_id.restype = ctypes.c_int
 rt.Index_NearestNeighbors_id.errcheck = check_return  # type: ignore
 
+try:
+    rt.Index_NearestNeighbors_id_v.argtypes = [
+        ctypes.c_void_p,
+        ctypes.c_int64,
+        ctypes.c_int64,
+        ctypes.c_uint32,
+        ctypes.c_uint64,
+        ctypes.c_uint64,
+        ctypes.c_uint64,
+        ctypes.c_void_p,
+        ctypes.c_void_p,
+        ctypes.c_void_p,
+        ctypes.c_void_p,
+        ctypes.c_void_p,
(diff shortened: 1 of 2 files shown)


hobu:
Looks like the code formatting nanny is failing on this.

FreddieWitherden:
I just looked over the code changes and all of the formatting seems fine.

hobu:
ruff-format is changing something, and the hook fails if of the nannies change anything. Maybe a false positive? 🤷

FreddieWitherden:
I think so?

Does the approach seem reasonable overall?  I am unsure why it does not fail when we provide objects to bulk-insert.  Maybe there is a check missing on the spatialindex side?

hobu:
It is just a complaint about code formatting as far as I understand. Logic and layout of PR looks fine to me.

FreddieWitherden:
Okay, I think we are okay to merge.

hobu:
Ruff is complaining about your quoting. Need to fix this or it will always complain

```
diff --git a/rtree/index.py b/rtree/index.py
index ef18743..92d51b8 100644
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -1539,7 +1539,7 @@ class Property:
                 setattr(self, k, v)
 
         # Consistency checks
-        if 'near_minimum_overlap_factor' not in state:
+        if "near_minimum_overlap_factor" not in state:
             nmof = self.near_minimum_overlap_factor
             ilc = min(self.index_capacity, self.leaf_capacity)
             if nmof >= ilc:

```

hobu:
did you mean to close this without merging?

FreddieWitherden:
Yes, now part of #342

PR #342, opened 2024-12-16
Add bulk query APIs

FreddieWitherden:
This is the counterpart to https://github.com/libspatialindex/libspatialindex/pull/268


FreddieWitherden:
Once we have a new libspatialindex release I'll get some tests written (at the moment they'll fail since the API is not yet merged) and do the docs.

PR #343, opened 2024-12-19
Update bulk interface to match changes in libspatialindex.

FreddieWitherden:


FreddieWitherden:
Are we okay to merge?

hobu:
can you please update the ticket title to describe *what* those changes in libspatialindex are and what their benefit/consequences might be? It will make it easier to roll up the release notes.

PR #344, opened 2025-01-04
rtree.index.Index: avoid use of args/kwargs

adamjstewart:
I spent longer than I would have liked trying to figure out why:
```python
index = Index(properties=Property(dimension=3, interleaved=False))
```
wasn't working. If we weren't using `*args, **kwargs` everywhere, I would have immediately figured this out. This PR starts with `Index`, but eventually I plan to move to other classes as well.

`*args, **kwargs` should be avoided for the following reasons:

* Does not report an issue when unknown kwargs are used (my problem)
* Does not report an issue when misspelled kwargs are used (actually discovered this in the tests)
* Does not support type hints

Basically the only time to use `*args, **kwargs` is when you have one function that wraps around another function and you need to pass through all arguments.

At the moment, this change is backwards incompatible since I replace `*args` with individual parameters and remove support for additional `Property` passthroughs. I could undo this change if we want to only remove `**kwargs` and keep things backwards compatible.

### DIFF
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -7,7 +7,7 @@ import pickle
 import pprint
 import warnings
 from collections.abc import Iterator, Sequence
-from typing import Any, Generator, Literal, overload
+from typing import Any, Literal, overload
 
 from . import core
 from .exceptions import RTreeError
@@ -196,7 +196,6 @@ class Index:
             raise RuntimeError(
                 "TPR-Tree type not supported with version of libspatialindex"
             )
-
 
         if filename:
             self.properties.storage = RT_Disk
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -7,7 +7,7 @@ import pickle
 import pprint
 import warnings
 from collections.abc import Iterator, Sequence
-from typing import Any, Literal, overload
+from typing import Any, Generator, Literal, overload
 
 from . import core
 from .exceptions import RTreeError
@@ -78,24 +78,23 @@ def _get_data(handle):
 class Index:
     """An R-Tree, MVR-Tree, or TPR-Tree indexing object"""
 
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
+    def __init__(
+        self,
+        filename: str | bytes | None = None,
+        stream: Any | None = None,
+        storage: ICustomStorage | None = None,
+        arrays: tuple | None = None,
+        interleaved: bool = True,
+        properties: Property | None = None,
+    ) -> None:
         """Creates a new index
 
-        :param filename:
-            The first argument in the constructor is assumed to be a filename
-            determining that a file-based storage for the index should be used.
-            If the first argument is not of type basestring, it is then assumed
-            to be an instance of ICustomStorage or derived class.
-            If the first argument is neither of type basestring nor an instance
-            of ICustomStorage, it is then assumed to be an input index item
-            stream.
+        :param filename: a filename determining that a file-based storage for the index
+            should be used.
 
-        :param stream:
-            If the first argument in the constructor is not of type basestring,
-            it is assumed to be an iterable stream of data that will raise a
-            StopIteration.  It must be in the form defined by the
-            :attr:`interleaved` attribute of the index. The following example
-            would assume :attr:`interleaved` is False::
+        :param stream: an iterable stream of data that will raise a StopIteration.
+            It must be in the form defined by the :attr:`interleaved` attribute of the
+            index. The following example would assume :attr:`interleaved` is False::
 
                 (id,
                  (minx, maxx, miny, maxy, minz, maxz, ..., ..., mink, maxk),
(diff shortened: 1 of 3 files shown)


rtree/index.py
```diff
@@ -78,24 +78,23 @@ def _get_data(handle):
 class Index:
     """An R-Tree, MVR-Tree, or TPR-Tree indexing object"""
 
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
+    def __init__(
+        self,
+        filename: str | bytes | None = None,
+        stream: Any | None = None,
```
adamjstewart, line 84: Haven't yet figured out how to type this

rtree/index.py
```diff
@@ -203,46 +197,17 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
             )
 
-        # interleaved True gives 'bbox' order.
-        self.interleaved = bool(kwargs.get("interleaved", True))
-
-        stream = None
-        arrays = None
-        basename = None
-        storage = None
-        if args:
```
adamjstewart, line 213: No more guessing games!

tests/test_index.py
```diff
@@ -849,12 +849,13 @@ def test_custom_storage_reopening(self) -> None:
...
         settings.buffering_capacity = 1
+        settings.overwrite = True
 
-        r1 = index.Index(storage, properties=settings, overwrite=True)
+        r1 = index.Index(storage=storage, properties=settings)
         r1.add(555, (2, 2))
         del r1
         self.assertTrue(storage.hasData)
 
-        r2 = index.Index(storage, properly=settings, overwrite=False)
```
adamjstewart, line 858: Actual typo/bug in our tests discovered by removing `**kwargs`

rtree/index.py
```diff
@@ -112,21 +111,15 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
                  object)
 
-        :param storage:
-            If the first argument in the constructor is an instance of
-            ICustomStorage then the given custom storage is used.
+        :param storage: A custom storage object to be used.
 
-        :param interleaved: True or False, defaults to True.
-            This parameter determines the coordinate order for all methods that
+        :param arrays: A tuple of arrays for bulk addition.
```
adamjstewart, line 116: Wasn't documented

tests/test_index.py
```diff
@@ -240,7 +240,7 @@ def test_intersection(self) -> None:
         self.assertTrue(0 in self.idx.intersection((0, 0, 60, 60)))
         hits = list(self.idx.intersection((0, 0, 60, 60)))
 
-        self.assertTrue(len(hits), 10)
+        self.assertEqual(len(hits), 10)
```
adamjstewart, line 243: Also a bug

adamjstewart:
Could use help figuring out the final remaining test failure...

hobu:
> Could use help figuring out the final remaining test failure...

Options ordering changed? The custom filename test failure must mean an unexpected default is getting used or something? 🤷

adamjstewart:
That's what I thought too, but that is just a simple test for `filename`, the first argument. I tried explicitly adding `filename=` but nothing changed.

adamjstewart:
First let's discuss whether or not this is something we want and later we can fix things.

hobu:
> something we want

Custom filenames? Yes, definitely. 

The janky kwargs/args configuration of the options? It was the style back in the day. I'm sure it could be remade to be something much better if someone has the time to put into it. I'm certainly not adverse to that, but it probably means making deprecating and then breaking things.

adamjstewart:
We can convert kwargs to parameters immediately without deprecation. Let me see if I can deprecate args and add parameters, it should actually be relatively easy.

adamjstewart:
Still need to deprecate (instead of remove) some properties and fix the test.

FreddieWitherden:
My preference here would be to create a subclass of Index which has a nicer and cleaner interface for construction.  It can then map this to the legacy API.  This would allow us to retain backwards compatibility and provide a route for code to be gradually updated.

Also, `Index` is not a great name for a class; `from rtree.index import Index` also looks a bit weird as an Index could really be anything (database Index, B tree, ...).

FreddieWitherden:
> @FreddieWitherden that would make it impossible to deprecate the old interface, because any instantiation of the subclass would also instantiate the parent class. The reverse would actually be better, new classes that implement the logic with the old class being a wrapper around those.

I'm not entirely sure what you mean?  The old interface absolutely can be deprecated and marked internal.  Then, after a suitable period of time it can be removed (and the code underpinning the new supported interface updated accordingly).  No code using the new interface would need to change (as how it is actually coded is very much an implementation detail so if it provides functionality itself or inherits it from a super class is immaterial to consumers). 

> However, I still think the current approach is easiest, as users will have the fewest things to change. My only remaining complaint about the current class is that it is highly dynamically typed, making type checking almost impossible. Any attempts to change that would likely result in a completely different library.

The problem with the current class is mostly around construction.  The actual query interface is okay.  Hence my suggestion around subclassing it and having these subclasses provide more reasonable constructors/static factory methods.

adamjstewart:
Are you suggesting to create a subclass of `Index` or an entirely new class that is independent of `Index`? If the former, we can't raise a warning message about deprecation in `Index` without also raising it in the subclasses. If the latter, this would be a lot of duplicated code, but I guess it would be fine in principle.

FreddieWitherden:
Rename `Index` to `_Index`.  Create a new class `Index(_Index)` and mark as deprecated.  Create a new class `Rtree(_Index)` (and friends) with a tweaked API that we want to support long term.

adamjstewart:
@hobu how would you like to proceed here? I think designing new subclasses is outside the scope of what I have time to work on. Could convert many of the args/kwargs to formal inputs though.

hobu:
> how would you like to proceed here? 

We can't break the old API. This library is way too old and entrenched and breaking things only for the sake of making them _better_ is going to annoy people and break downstream users who might not even know they're using the library.

We can make a new API and replumb the old API to take advantage of new internals, however (anyone want to jump to pybind11 or similar while we're at? 😄). The amount of work to do this is what prevents anyone from starting, unfortunately.

FreddieWitherden:
So I think we should get a release out the door and then decide how far we want to go in terms of a new API.

The limiting factor for me at the moment is on the libspatialindex side where I feel the MVR tree is not particularly useful in its current form and the TPR tree really is quite a different beast entirely.

hobu:
What's left to do before we cut a release?

FreddieWitherden:
#346 would be nice to have, but I am unsure how to handle the type annotations properly for the new interfaces.

adamjstewart:
This PR has turned out to be more work than expected to ensure backwards compatibility. TorchGeo is also planning on moving away from R-tree in the future, so I don't want to spend too much time on this. I'll close for now, but if anyone wants to revive this to at least get rid of kwargs, I would be happy to review.

hobu:
> TorchGeo is also planning on moving away from R-tree in the future

I'm curious what prompted this. Need for pure Python?

PR #345, opened 2025-01-06
rtree.index: fix test bugs

adamjstewart:
Fixes a few bugs discovered in #344. Separated them out into their own PR to make them easier to review/merge.

### DIFF
--- a/tests/test_index.py
+++ b/tests/test_index.py
@@ -846,15 +846,11 @@ class IndexCustomStorage(unittest.TestCase):
         """Reopening custom index storage works as expected"""
 
         storage = DictStorage()
-        settings = index.Property()
-        settings.writethrough = True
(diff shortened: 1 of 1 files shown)


tests/test_index.py
```diff
@@ -855,6 +855,6 @@ def test_custom_storage_reopening(self) -> None:
         del r1
         self.assertTrue(storage.hasData)
 
-        r2 = index.Index(storage, properly=settings, overwrite=False)
+        r2 = index.Index(storage, properties=settings, overwrite=False)
```
mwtoews, line 858: I suppose that the default properties were sufficient for this test to succeed. Maybe we should remove the properties for this test? It would remove L849-L851 and adjust this line and L853.

mwtoews (approved):

PR #346, opened 2025-01-06
Document bulk interfaces

FreddieWitherden:


adamjstewart:
Would also be nice to add type hints

PR #358, opened 2025-03-15
Fixing an incorrect reassignment in nearest_v and intersection_v

Atilleusz:
The existing implementation confuses `numpy.resize` that returns a new resized array, with `numpy.ndarray.resize` that resizes the array in-place and returns None. Relevant numpy documentation:

https://numpy.org/doc/stable/reference/generated/numpy.resize.html
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.resize.html

This causes calls to nearest_v and intersection_v to fail in cases where `nr.value == n - offn` evaluates to false.

### DIFF
--- a/rtree/index.py
+++ b/rtree/index.py
@@ -1103,7 +1103,7 @@ class Index:
                 offi += counts[offn : offn + nr.value].sum()
                 offn += nr.value
 
-                ids = ids.resize(2 * len(ids), refcheck=False)
+                ids.resize(2 * len(ids), refcheck=False)
 
     def nearest_v(
         self,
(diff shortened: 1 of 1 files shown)


mwtoews:
Thanks! We'll need to add a test. Do you have an example handy?

Atilleusz:
```
import numpy as np
import rtree

mins = np.vstack([(0.0, 0.0), (5.0, 5.0)], dtype=float)
maxs = np.vstack([(10.0, 10.0), (15.0, 15.0)], dtype=float)
props = rtree.index.Property(
    dimension=2,
    storage=rtree.index.RT_Memory,
)

idx = rtree.index.Index(properties=props, interleaved=False)
idx.insert(1, (0, 100, 0, 100))
idx.insert(2, (7.5, 7.5, 7.5, 7.5))
idx.insert(3, (2.5, 2.5, 2.5, 2.5))
idx.intersection_v(mins, maxs)
```
I managed to reproduce the issue with this code. (A bit of work because in my situation it got to the code-path in a convoluted way)

numpy version: 2.2.3
rtree version: 1.4.0

mwtoews:
Perfect, thanks! It seems that critical tests for #342 (and related) were missing from the 1.4.0 release. I guess I'll pull together some more tests with possible fixes for a 1.4.1 release.

mwtoews (approved):
I'll address the tests in a followup.

PR #360, opened 2025-03-24
Add spatialindex version to tests, add common pytest configuration

mwtoews:
This adds a `common` module for the tests, for items that aren't fixtures. So far it gets the libspatialindex version string and tuple.

Also establish a pytest configuration.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,8 @@ name: Build and upload to PyPI
 on:
   workflow_dispatch:
   pull_request:
+    paths:
+      - '.github/workflows/deploy.yml'
   push:
     branches:
       - main
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   conda:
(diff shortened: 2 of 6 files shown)


adamjstewart:
Wouldn't it make more sense to define this as a reusable fixture in a `conftest.py` file? https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files

mwtoews:
Aren't fixtures specific to running pytests? I.e., shouldn't `sidx_version` be just a regular variable? I generally keep `conftest.py` only for fixtures, not other variables.

Also, it seems that I don't know how to stop `deploy` from always running, despite adding a "paths" filter for `pull_requests`.

mwtoews:
I can't see an easy way to leverage fixtures as simple variables (since fixtures are special functions), so will proceed with a simple approach in tests for now.

For 1.5.x, both `sidx_version_string` and `sidx_version` cam be moved to the main module, as these could be handy for users too.

Just noticed that the conda tests were always excluding macos and SIDX 1.8.5, which is fixed to cover these.

PR #362, opened 2025-03-29
Minor refactor of code blocks in docs

mwtoews:
This is a relatively minor refactor of the code blocks in the docs. Mostly to update the look-and-feel via blacken-docs (not added as a check or pre-commit), but also to check that they are running as expected, now checked in unit tests.

There were some Python 2 left-overs updated for Python 3. Also, temporary directories are used for the disk serialization examples, so they don't pollute the current working directories (and these tests break when run more than once due to the append mode).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
       run: pip install -e .
 
     - name: Test with pytest
-      run: pytest --import-mode=importlib -Werror -v --doctest-modules rtree tests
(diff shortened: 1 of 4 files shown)


mwtoews:
This effort does not attempt to refactor the module docs or examples. I'll consider re-writing the docs using a numpy-style for 1.5, which may also update some of these examples (blacken docs doesn't work on examples in `*.py` files).

adamjstewart (approved):

PR #367, opened 2025-05-11
Resolve some issues in the batch API

FreddieWitherden:
This elides some unnecessary copies of NumPy arrays when using the batch interface.


FreddieWitherden:
Test failures appear to be problems with the test themselves (relying on undefined behaviour for the ordering of `os.listdir()`).

FreddieWitherden:
@hobu Let me know what you think

FreddieWitherden:
Tests should be fixed.

FreddieWitherden:
@hobu I've added some code to handle mis-aligned arrays.  Certainly an edge-case but something which we should handle nevertheless.

FreddieWitherden:
@hobu I think with this we are okay to release v1.4.1?

hobu:
1.4.1 was released two days ago! Let me know if you still have any issues.

PR #377, opened 2025-08-10
CI Refactor

hobu:
This moves all CI operations into a single multi-step workflow. The primary reason for this was to take advantage of [the attestation](https://github.blog/news-insights/product-news/introducing-artifact-attestations-now-in-public-beta/) capabilities of GitHub.


hobu:
Other reasons for bringing things into a single workflow:
* Security of passing artifact access between distinct workflows is painful
* We don't run these often and stuff rots over time – easier to find things in one place
* Nice flowchart view

PR #387, opened 2025-09-01
Test with Python 3.14 as upper version

mwtoews:
Start testing with Python 3.14, expected in October 2025 ([PEP 745](https://peps.python.org/pep-0745/)).

The `ubuntu` CI jobs can be pruned down to the upper and lower Python version with no loss of coverage. Full range of Python versions are tested by cibuildwheel with tox.

Also add a new classifier. (Although, I could also be convinced to remove most of these classifiers, as they are not always relevant, i.e. previous releases on PyPI could usually work with with future versions of Python).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         # test oldest and newest versions of python and libspatialindex
-        python-version: ['3.9', '3.13']
+        python-version: ['3.9', '3.14']
         sidx-version: ['1.8.5', '2.1.0']
         exclude:
           - os: 'macos-latest'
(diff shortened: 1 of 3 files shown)


mwtoews:
Probably easiest to hold this one until a stable Python 3.14 is in conda-forge (rather than fiddling around with various release-candidate channels).

hobu:
> Probably easiest to hold this one until a stable Python 3.14 is in conda-forge 

Yeah, this should be available in a few weeks. There's tons of 314 churn in the feedstocks right now.

.github/workflows/test.yml
```diff
@@ -58,7 +58,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
```
adamjstewart: I would personally prefer testing every minor version. Especially for a library like this that tightly integrates with CPython, there may be differences in minor versions we want to test. We can drop support for 3.9 or other versions if you're concerned about the total number of tests.

PR #404, opened 2026-04-01
fix: add fallback for finding bundled library when importlib.metadata unavailable

Wheest:
## Summary

When running in sandboxed environments like Bazel (with `rules_python`), `importlib.metadata.files()` may return `None` or fail to locate files correctly, even though the bundled `libspatialindex` library exists in the wheel.

This PR adds a fallback that searches for the bundled library relative to the package directory (`rtree.libs/`), which works regardless of metadata availability.

## Problem

In Bazel's sandboxed test environment:
```
OSError: Could not load libspatialindex_c library
```

The library exists at `rtree.libs/libspatialindex-*.so` but `importlib.metadata.files("rtree")` returns `None` because Bazel doesn't properly expose package metadata.

## Solution

After the existing `importlib.metadata` lookup, add a simple fallback:
```python
libs_dir = _cwd.parent / "rtree.libs"
if libs_dir.exists():
    for so_file in libs_dir.glob("libspatialindex*.so"):
        _candidates.insert(1, so_file)
        break
```

This uses the already-defined `_cwd` (`Path(__file__).parent`) to locate the library relative to the package, which works in any environment.

## Testing

- Added test that mocks `importlib.metadata.files()` returning `None`
- Verified fix works in Bazel environment (commit 1 fails, commit 2 passes)

### DIFF
--- a/rtree/finder.py
+++ b/rtree/finder.py
@@ -86,6 +86,14 @@ def load() -> ctypes.CDLL:
             except importlib.metadata.PackageNotFoundError:
                 pass
 
+            # Fallback: search relative to package directory
+            # (works when importlib.metadata is unavailable, e.g. Bazel, some venvs)
+            libs_dir = _cwd.parent / "rtree.libs"
+            if libs_dir.exists():
+                for so_file in libs_dir.glob("libspatialindex*.so"):
+                    _candidates.insert(1, so_file)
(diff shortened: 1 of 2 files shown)


mwtoews (approved):
Thanks for fix with test, looks good.

PR #412, opened 2026-06-22
Address pytest warning about instance attributes with a classmethod

hobu:
* Drop Python 3.9 testing

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         # test oldest and newest versions of python and libspatialindex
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         # test oldest and newest versions of python and libspatialindex
-        python-version: ['3.9', '3.14']
+        python-version: ['3.10', '3.14']
```
mwtoews, line 28: Dropping Python support should be done in a separate PR, as to not mix-up intents. It looks like the [previous test with Python 3.9 ran fine](https://github.com/Toblerity/rtree/actions/runs/27546409217/job/81420839607).

PR #420, opened 2026-09-05
Security hardening: remove use of pickle

adamjstewart:
Discussed via private security report @hobu

`pickle.loads` on unvalidated inputs is susceptible to remote code execution. This PR replaces pickling with simple bytes/str conversion. The main downside of this is that for users storing things other than str, anything involving serialization/deserialization will change their objects to str. For example, the previous usage of rtree in TorchGeo would probably break. This involved changing far more unit tests than expected, which I don't love.

Alternatives include json (which would probably support all objects in unit tests) or adding a new dependency for safer pickling.

Note that this is a backwards-incompatible change and should probably warrant at least a minor version bump.


adamjstewart:
FWIW, I also have a json version of this that only changes a couple tests (tuples become lists).

hobu (approved):

hobu:
@mwtoews do you have an opinion on this? It would definitely cause breakage for those using clustered index storage. I'm not sure how common its use was.

mwtoews:
What about handling objects with `int` types specially like this?
```python
    def dumps(self, obj: object) -> bytes:
        if isinstance(obj, int):
            return bytes("int:" + str(obj), "utf-8")
        return bytes("obj:" + str(obj), "utf-8")

    def loads(self, string: bytes) -> str | int:
        string_ = str(string, "utf-8")
        if string_.startswith("int:"):
            return int(string_[4:])
        elif string_.startswith("obj:"):
            return string_[4:]
        raise NotImplementedError("Custom loads() not implemented for this type")
```
this could reduce the number of changes / possible breakage. Just throwing ideas around, feedback welcome.

---

On the topic of JSON serialization, there is a small example here:
https://github.com/Toblerity/rtree/blob/cb379b231037c2e78cbc9d35b12984fbb3e506ff/rtree/index.py#L173-L194
(This is back from 54c78bba54431f618392531a1b647d7c7f16c175 when `simplejson` was used instead of the `json` built-in)

adamjstewart:
I guess my first question is why stop there? Float seems like a good idea, and then why not all built in types? Curious if simplejson has the same security considerations as json.

adamjstewart:
Let's assume that simplejson has the same concerns as json, that it doesn't allow RCE but it does allow DoS. Do we:

1. manually parse all builtin types we care about as str,
2. switch to json and document the DoS vulnerability, or
3. keep pickle and document the RCE vulnerability?

hobu:
My preference would be for 1 – we support basic plain old data types and expand that out to `json` support if enough complaints arise post-release. The entire approach of storing pickles alongside index entries hasn't worked so well for a long time anyway https://github.com/Toblerity/rtree/pull/197 

If the data being stored is anything but plain old data types, the implementation should account for attempting to open up the index entries that do have pickles in them (throw an exception with the pickle data captured?). If the user wanted to de-pickle them themselves, they should be able to do that, but Rtree should no longer provide that as a convenience.

adamjstewart:
Okay, which builtin types do we want to support? Only str? Int? Float? Others?

mwtoews:
Probably the basic ones: `int`, `str` and `float`. It gets trickier with (e.g) `tuple`. And for `float`, use [`float.hex()` / `fromhex()`](https://docs.python.org/3/library/stdtypes.html#float.hex) methods to serialize exact representations.

adamjstewart:
Other builtin types include bool, complex, bytes, memoryview. The rest seem to be nested types, which I agree are too much work to support. Will implement the bare minimum until someone requests more.

mwtoews:
Thanks @adamjstewart for starting this. I have some PR feedback in the form of a commit. I've had to do more strict type matching to avoid inheritance things like `isinstance(True, int)`. Also add some tests. Please revise!

adamjstewart:
Thanks @mwtoews, changes look good to me!

mwtoews (approved):
Feel free to make any other minor changes, and merge when you are happy.

rtree/index.py
```diff
@@ -329,10 +328,28 @@ def __setstate__(self, state: dict[str, Any]) -> None:
         self.handle = IndexHandle(self.properties.handle)
 
     def dumps(self, obj: object) -> bytes:
-        return pickle.dumps(obj)
-
-    def loads(self, string: bytes) -> object:
-        return pickle.loads(string)
+        obj_type = type(obj)
+        # use strict type() matching rather than isinstance()
+        if obj_type is bool or obj is None:
```
hobu, line 333: oh, I see what you guys were talking about in regard to JSON.  As currently implemented, this is a kind of custom format, right? Pickle was too, of course. What's the case against using JSON + types as the serialization here instead of this approach?

adamjstewart: Only case against is the following warning from https://docs.python.org/3/library/json.html:

> [!WARNING]
> Be cautious when parsing JSON data from untrusted sources. A malicious JSON string may cause the decoder to consume considerable CPU and memory resources. Limiting the size of data to be parsed is recommended.

If my memory serves me correctly, this allows for DoS-style attacks, so it's medium-severity instead of high-severity. But if we manually limit the size of the object we are de-JSONing, we may be able to avoid that. Let me open a separate PR to see what that would look like, I think it would be much simpler and support a larger range of nested types.

hobu:
closing this in favor of #423

PR #423, opened 2026-09-17
Security hardening: pickle -> json

adamjstewart:
Alternative to #420 

@hobu @mwtoews I think I found a way to make json safe. This supports a much wider variety of types than manual str parsing, and should be unnoticeable for most users. The only difference found during testing is that tuples are replaced by lists, but that may not matter for most use cases. Arbitrary custom objects are no longer supported, but we could document this. If this looks good I can also add unit tests for this new error path.

Closes #420


rtree/index.py
```diff
@@ -329,10 +329,18 @@ def __setstate__(self, state: dict[str, Any]) -> None:
         self.handle = IndexHandle(self.properties.handle)
 
     def dumps(self, obj: object) -> bytes:
-        return pickle.dumps(obj)
-
-    def loads(self, string: bytes) -> object:
-        return pickle.loads(string)
+        return bytes(json.dumps(obj), "utf-8")
```
hobu: We should also guard the production side to limit less than 1024 bytes of JSON if we're guarding the consumption side.

rtree/index.py
```diff
@@ -328,11 +329,22 @@ def __setstate__(self, state: dict[str, Any]) -> None:
...
         self.handle = IndexHandle(self.properties.handle)
 
+    # https://docs.python.org/3/library/json.html
+    #
+    # Be cautious when parsing JSON data from untrusted sources. A malicious JSON
+    # string may cause the decoder to consume considerable CPU and memory resources.
+    # Limiting the size of data to be parsed is recommended.
     def dumps(self, obj: object) -> bytes:
-        return pickle.dumps(obj)
+        if sys.getsizeof(obj) < 1024:
```
adamjstewart: Could also check length of encoded str, the above warning appears to only apply to the decoder, not the encoder.

rtree/index.py
```diff
@@ -328,11 +329,22 @@ def __setstate__(self, state: dict[str, Any]) -> None:
         self.__dict__.update(state)
         self.handle = IndexHandle(self.properties.handle)
 
+    # https://docs.python.org/3/library/json.html
+    #
+    # Be cautious when parsing JSON data from untrusted sources. A malicious JSON
+    # string may cause the decoder to consume considerable CPU and memory resources.
+    # Limiting the size of data to be parsed is recommended.
```
hobu, line 338:
```suggestion
    # Limit the object to only 1k of JSON serialization.
```

rtree/index.py
```diff
@@ -1,10 +1,11 @@
...
 import ctypes
+import json
 import os
 import os.path
-import pickle
 import pprint
+import sys
 import warnings
 from collections.abc import Iterator, Sequence
 from typing import Any, Literal, overload
```
hobu, line 11:
```suggestion
from typing import Any, Literal, overload

INDEX_JSON_SERIALIZATION_LIMIT_SIZE = 1024
```

rtree/index.py
hobu:
```suggestion
        if sys.getsizeof(obj) < INDEX_JSON_SERIALIZATION_LIMIT_SIZE:
```

rtree/index.py
```diff
@@ -328,11 +329,22 @@ def __setstate__(self, state: dict[str, Any]) -> None:
...
-        return pickle.dumps(obj)
+        if sys.getsizeof(obj) < 1024:
+            return bytes(json.dumps(obj), "utf-8")
+        else:
+            raise TimeoutError("Object is too large to quickly encode")
 
-    def loads(self, string: bytes) -> object:
-        return pickle.loads(string)
+    def loads(self, string: bytes) -> Any:
+        if len(string) < 1024:
```
hobu:
```suggestion
        if len(string) < INDEX_JSON_SERIALIZATION_LIMIT_SIZE:
```

rtree/index.py
```diff
@@ -328,11 +329,22 @@ def __setstate__(self, state: dict[str, Any]) -> None:
...
+            return json.loads(str(string, "utf-8"))
+        else:
+            raise TimeoutError("JSON string is too large to quickly decode")
```
hobu:
```suggestion
            raise TimeoutError("JSON string is too large to quickly decode")
            raise TimeoutError(f"Unable to load serialized index data. The JSON string is above the serialization limit of {INDEX_JSON_SERIALIZATION_LIMIT_SIZE")
```

hobu (changes requested):
I don't know if we should let users dynamically control this size limitation, but we should parameterize it to make it convenient to change.

rtree/index.py
adamjstewart: I wouldn't change this, this is just a direct quote from the docs

adamjstewart:
Reminder: double check for other places in the docs that mention pickle and document that only valid JSON objects are serializable.

hobu (approved):

